PR #17, opened 2015-12-01
Fix "redefinition of Base.read" warning

maximsch2:
- move Base.read redefinition to io.jl
- update it based on Julia source code,
- only use it when VERSION < 0.4.1


quinnj:
the versioning here isn't necessary because CSV already requires julia 0.4.1.  we do, however, need to address the `Base.read` problem.

PR #20, opened 2016-05-20
Allow specifying different floating point types

ScottPJones:
Fix bug with parsing BigFloat & DecFP (and possibly others) with workaround

This allows a keyword argument `floattype`, that can accept any type that is `<: AbstractFloat`, so that you can load a CSV file using decimal floating point (such as with `DecFP.jl`, higher precision floating point (with `BigFloat`), or even things like `Unums`.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -36,6 +36,7 @@ Represents the various configuration settings for csv file parsing.
  * `escapechar`::Union{Char,UInt8} = the character that escapes a `quotechar` in a quoted field
  * `null`::String = the ascii string that indicates how NULL values are represented in the dataset
  * `dateformat`::Union{AbstractString,Dates.DateFormat} = how dates/datetimes are represented in the dataset
+ * `floattype`::DataType = type used (must be <: AbstractFloat) to represent floating point numbers
 """
 type Options
     delim::UInt8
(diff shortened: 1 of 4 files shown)


quinnj:
Let me simmer on this for a few days and think it out. I think the approach here is ok (gets the job done), but I think we might be able to simplify without having to modify so much machinery.

ScottPJones:
Consider this WIP then, on my way home from ODSC today, I realized it should probably allow setting the default types for strings and integers also (and be able to handle setting the default setting for integers to an `AbstractFloat` type as well).  I also realized there is a problem currently if somebody wanted to get the strings as something like `UTF16String` (or even a multibyte character set that doesn't guarantee that subsequent bytes in a character are all > 0x7f like UTF-8, as somebody could allow that in the future, if we had a really nice string package along the lines of combining `Strings.jl` and `StringEncodings.jl` ;-) )  What you've done with `CSV.jl` is very nice - let me know if you come up with a better approach to handling this and would like me to implement it.

ScottPJones:
It looks like there are some bugs related to v0.5 changes in the tests themselves, causing the tests to be broken.
I'll put fixes for those in another PR today if possible.

quinnj:
Yeah, we need some additional Compat entries for the new `bytestring` deprecations. See https://github.com/JuliaLang/Compat.jl/pull/203

ScottPJones:
I've also been wondering about an easier way of describing the types of a column, maybe as a convenience function wrapping the CSV constructor.
It may be a crazy thought, but I was thinking of possibly a string, that could be used with an optional `Dict{Char,DataType}`.
Let's say you had 4 columns, the 1st a string, 2nd an integer, 3rd a float, 4th a date.
The string "sifd" could describe that.  If, for example, you wanted a non-default type for floats, you could pass a dict with 'f'=>Dec64, for example. A space might be used to indicate that the type for the column would be inferred as is possible now.  The user could actually define whatever characters they want, to select what type should be used (say, to allow having one column being `Float64` and another being `Dec64`.

quinnj:
Fixed by 692fb208b0a08fd00779ab25d307c8822c0d3394. Now you can just do `CSV.read(file; types=[Dec64])` for a single Dec64 column and it will "just work".

PR #22, opened 2016-05-31
fix String compat with v0.4

felipenoris:
This PR is related to https://github.com/JuliaDB/DataStreams.jl/pull/9.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -7,7 +7,7 @@ const PointerString = Data.PointerString
 const NULLSTRING = Data.NULLSTRING
 
 immutable CSVError <: Exception
-    msg::@compat(String)
+    msg::Compat.UTF8String
 end
 
 const RETURN  = UInt8('\r')
@@ -34,7 +34,7 @@ Represents the various configuration settings for csv file parsing.
  * `delim`::Union{Char,UInt8} = how fields in the file are delimited
  * `quotechar`::Union{Char,UInt8} = the character that indicates a quoted field that may contain the `delim` or newlines
  * `escapechar`::Union{Char,UInt8} = the character that escapes a `quotechar` in a quoted field
- * `null`::String = the ascii string that indicates how NULL values are represented in the dataset
+ * `null`::String = indicates how NULL values are represented in the dataset
  * `dateformat`::Union{AbstractString,Dates.DateFormat} = how dates/datetimes are represented in the dataset
 """
 type Options
@@ -43,13 +43,13 @@ type Options
     escapechar::UInt8
     separator::UInt8
     decimal::UInt8
-    null::@compat(String) # how null is represented in the dataset
+    null::Compat.UTF8String # how null is represented in the dataset
     nullcheck::Bool   # do we have a custom null value to check for
     dateformat::Dates.DateFormat
     datecheck::Bool   # do we have a custom dateformat to check for
 end
 
-Options(;delim=COMMA,quotechar=QUOTE,escapechar=ESCAPE,null::@compat(String)="",dateformat=Dates.ISODateFormat) =
+Options(;delim=COMMA,quotechar=QUOTE,escapechar=ESCAPE,null="",dateformat=Dates.ISODateFormat) =
     Options(delim%UInt8,quotechar%UInt8,escapechar%UInt8,COMMA,PERIOD,
             null,null != "",isa(dateformat,Dates.DateFormat) ? dateformat : Dates.DateFormat(dateformat),dateformat == Dates.ISODateFormat)
 function Base.show(io::IO,op::Options)
@@ -67,7 +67,7 @@ type Source <: Data.Source
     options::Options
     data::UnsafeBuffer
     datapos::Int # the position in the IOBuffer where the rows of data begins
-    fullpath::@compat(String)
+    fullpath::Compat.UTF8String
 end
 
 function Base.show(io::IO,f::Source)
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -15,7 +15,7 @@ constructs a `CSV.Source` file ready to start parsing data from
 * `delim`::Union{Char,UInt8} = how fields in the file are delimited
 * `quotechar`::Union{Char,UInt8} = the character that indicates a quoted field that may contain the `delim` or newlines
 * `escapechar`::Union{Char,UInt8} = the character that escapes a `quotechar` in a quoted field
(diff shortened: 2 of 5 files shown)


src/CSV.jl
```diff
@@ -43,13 +43,13 @@ type Options
     escapechar::UInt8
     separator::UInt8
     decimal::UInt8
-    null::@compat(String) # how null is represented in the dataset
+    null::Compat.ASCIIString # how null is represented in the dataset
```
nalimilan: Why not allow UTF-8 here too?

src/getfields.jl
```diff
@@ -278,8 +278,16 @@ function parsefield(io::Union{IOBuffer,UnsafeBuffer}, ::Type{DateTime}, opt::CSV
...
             end
         end
-        val = String(pointer(io.data)+pos, position(io)-pos-quoted-1+end_of_file)
+        val = pointer_to_string(pointer(io.data)+pos, position(io)-pos-quoted-1+end_of_file)
         val == opt.null && return DateTime(0,1,1), true
         return DateTime(val, opt.dateformat)::DateTime, false
     end
 end
+
+@generated function pointer_to_string(ptr, len)
```
nalimilan: Should be added to Compat.jl.

test/runtests.jl
```diff
@@ -20,7 +15,7 @@ f = CSV.Source(dir * "test_utf8.csv")
 @test f.options.delim == UInt8(',')
 @test f.schema.cols == 3
 @test f.schema.rows == 3
-@test f.schema.header == ["col1","col2","col3"]
+@test f.schema.header == Compat.UTF8String["col1","col2","col3"]
```
nalimilan: Shouldn't be needed AFAIK. The string type does not affect `==`.

test/runtests.jl
```diff
@@ -1,15 +1,10 @@
...
-### File.jl
-
-# dir = "/Users/jacobquinn/.julia/v0.4/CSV/test/test_files/"
 dir = joinpath(dirname(@__FILE__),"test_files/")
 
-# test 0.3 and 0.4
-
 #test on non-existent file
-@test_throws ArgumentError CSV.Source("");
+@test_throws ArgumentError CSV.Source(Compat.UTF8String(""));
```
nalimilan: While you're at it, it would be cool to make `CSV.Source` accept any `AsbtractString`, and convert it to `UTF8String` automatically.

test/runtests.jl
nalimilan: Same for all cases below.

src/CSV.jl
```diff
@@ -43,13 +43,13 @@ type Options
...
     nullcheck::Bool   # do we have a custom null value to check for
     dateformat::Dates.DateFormat
     datecheck::Bool   # do we have a custom dateformat to check for
 end
 
-Options(;delim=COMMA,quotechar=QUOTE,escapechar=ESCAPE,null::@compat(String)="",dateformat=Dates.ISODateFormat) =
+Options(;delim=COMMA,quotechar=QUOTE,escapechar=ESCAPE,null::Compat.ASCIIString="",dateformat=Dates.ISODateFormat) =
```
nalimilan: Given that no other argument has a type here, just remove it and leave the constructor do the conversion automatically.

src/getfields.jl
nalimilan: Actually, I think you want `bytestring` here, which is in the latest version of Compat.

felipenoris:
Thanks, @nalimilan .

src/Source.jl
```diff
@@ -121,20 +121,20 @@ function Source(;fullpath::Union{AbstractString,IO}="",
             datapos = position(source)
             cols = length(CSV.readsplitline(source,options.delim,options.quotechar,options.escapechar))
             seek(source, datapos)
-            columnnames = @compat(String)["Column$i" for i = 1:cols]
+            columnnames = Compat.UTF8String["Column$i" for i = 1:cols]
```
nalimilan: What happens if you remove the element type here and below?

felipenoris:
@nalimilan , I guess it works!

nalimilan:
> @nalimilan , I guess it works!

So why not remove all of these?

felipenoris:
@nalimilan I did it already. Or did I miss something?

test/runtests.jl
```diff
@@ -64,7 +59,7 @@ f = CSV.Source(dir * "test_single_column.csv")
 @test position(f.data) == 5
 @test f.schema.cols == 1
 @test f.schema.rows == 3
-@test f.schema.header == ["col1"]
+@test f.schema.header == Compat.UTF8String["col1"]
```
nalimilan: I meant this one as well as similar cases.

felipenoris:
@nalimilan , my bad. I fixed only `src/Source.jl`. I'll check for similar cases.

felipenoris:
@nalimilan , done.

test/runtests.jl
```diff
@@ -162,10 +157,10 @@ ds = Data.stream!(f, Data.Table)
 @test ds[1,1].value == 1.0
 @test string(ds[1,2].value) == "2.0"
 @test string(ds[2,2].value) == "NULL"
-f = CSV.Source(dir * "test_missing_value_NULL.csv";null="NULL")
+f = CSV.Source(dir * "test_missing_value_NULL.csv";null=Compat.UTF8String("NULL"))
```
nalimilan: Do you also need this one and the two other cases below? It would be nice if `Source` converted its argument to the correct string type.

Other than that, looks good to me.

src/getfields.jl
```diff
@@ -214,7 +214,7 @@ function parsefield(io::Union{IOBuffer,UnsafeBuffer}, ::Type{Date}, opt::CSV.Opt
                 break
             end
         end
-        val = String(pointer(io.data)+pos, position(io)-pos-quoted-1+end_of_file)
+        val = Compat.bytestring(pointer(io.data)+pos, position(io)-pos-quoted-1+end_of_file)
```
nalimilan: Sorry, I think I was mistaken. According to Compat's README, this should be `@compat String(...)`, and same below.

felipenoris:
I removed a few more `Compat.UTF8String` from `runtests.jl`. I guess conversion is already working.
`@compat String(...)` does nothing for me. So I kept `Compat.bytestring(...)`.

quinnj:
Thanks @felipenoris and @nalimilan!

Issue #49, opened 2016-11-29
Problem parsing null non-string fields in tab-separated file


joshbode:
I'm having trouble parsing tab-separated files with non-string columns (as inferred from the data or explicitly declared via the `types` keyword) containing null values (denoted by adjacent delimiters). The parser is grabbing the following cell, which can be seen in the following example (Julia 0.5, CSV.jl 0.1.1):

## Input:
```bash
$ cat /tmp/foo.txt
A	B	C	D
1	2016	x	100
2		2014	200
```

## Result:
```julia
julia> CSV.read("/tmp/foo.txt"; delim='\t')
2×4 DataFrames.DataFrame
│ Row │ A │ B    │ C     │ D     │
│ 1   │ 1 │ 2016 │ "x"   │ 100   │
│ 2   │ 2 │ 2014 │ "200" │ #NULL │
```

## Expected:
```
│ Row │ A │ B     │ C      │ D   │
│ 1   │ 1 │ 2016  │ "x"    │ 100 │
│ 2   │ 2 │ #NULL │ "2014" │ 200 │
```
If column C in row 2 did not contain a value of the same inferred type as column B of row 1 (Int64) then the parser would fail, unable to parse the cell as an integer, e.g.

## Input:
```bash
$ cat /tmp/bar.txt
A	B	C	D
1	2016	x	100
2		y	200
```

## Result:
```julia
julia> CSV.read("/tmp/bar.txt"; delim='\t')
ERROR: CSV.CSVError("error parsing a `Int64` value on column 2, row 2; encountered 'y'")
 in checknullend at /Users/josh/.julia/v0.5/CSV/src/parsefields.jl:56 [inlined]
 in parsefield at /Users/josh/.julia/v0.5/CSV/src/parsefields.jl:127 [inlined]
 in parsefield at /Users/josh/.julia/v0.5/CSV/src/parsefields.jl:107 [inlined]
 in streamfrom(::CSV.Source, ::Type{DataStreams.Data.Field}, ::Type{Nullable{Int64}}, ::Int64, ::Int64) at /Users/josh/.julia/v0.5/CSV/src/Source.jl:185
 in streamto!(::DataFrames.DataFrame, ::Type{DataStreams.Data.Field}, ::CSV.Source, ::Type{Nullable{Int64}}, ::Type{Nullable{Int64}}, ::Int64, ::Int64, ::Data
[...]

PR #50, opened 2016-11-29
Fix for issue #49: Null-delimiter parsing

joshbode:
Addresses issue #49 by checking current value is not the delimiter (e.g. tab or space).


quinnj:
Windows failures are un-related. I need to fix that anyway. Thanks!

PR #56, opened 2017-01-09
Don't require write permissions when reading a file

andreasnoack:
Similar to https://github.com/JuliaLang/julia/pull/19337

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -58,7 +58,10 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         fs = nb_available(fullpath)
         fullpath = isdefined(fullpath, :name) ? fullpath.name : "__IO__"
     else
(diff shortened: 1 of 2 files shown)


andreasnoack:
Tests should pass with #57

src/Source.jl
```diff
@@ -58,7 +58,10 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         fs = nb_available(fullpath)
         fullpath = isdefined(fullpath, :name) ? fullpath.name : "__IO__"
     else
-        source = IOBuffer(use_mmap ? Mmap.mmap(fullpath) : open(Base.read, fullpath))
+        # source = IOBuffer(use_mmap ? Mmap.mmap(fullpath) : open(Base.read, fullpath))
```
quinnj: can you remove the commented out code here?

quinnj: nm, I'll do this myself locally with some other changes.

PR #65, opened 2017-03-13
Refactor readsplitline()

### LINKED ISSUE #64: Types and values of quoted strings
AFAIU automatic type detection does not take into account whether the extracted string was quoted or not, which leads to incorrect predictions for the column containing quoted numbers as the first N elements. 

Checking for quotes should also be done when checking for `null` strings: quoted string should be automatically regarded as non-null.
 
Actually, encountered this while trying to convert `RDatasets` into using `DataTables` (`datasets/attenu.csv.gz`). `RDataset` has a nice test set of .csv files.

alyst:
Motivated by #64, I've refactored `readsplitline()`:
* `RawField` stores the unparsed string contents of the field + `isquoted` flag
* `readsplitline!()` reuses a vector of field contents to reduce the memory footprint
* `readsplitline!()` is now aware of its current parsing state; although the function became somewhat bulkier, that allowed catching some misformated CSV lines (mupliple quoted strings in one field, missing quote etc) + tests
* `detecttype()` treats all quoted fields as a string (fixes #64) + test

Also, `Sink` doesn't quote/escape the `options.null` string (otherwise the corresponding column type would considered a string).


alyst:
0.5 tests pass, the nightly CI failures look unrelated.

quinnj:
Sorry for the delay on this, it's a busy time getting ready for 0.6. Thanks for the contribution!

PR #70, opened 2017-03-22
Fix CSV.read error on an empty CSV with header specified in keyword argument

### LINKED ISSUE #69: `CSV.read` on empty buffer fails when `header::Vector{String}` is specified
These examples behave identically, as expected:

```julia
julia> CSV.read(IOBuffer("a,b,c\n1,2,3\n4,5,6"))
2×3 DataFrames.DataFrame
│ Row │ a │ b │ c │
├─────┼───┼───┼───┤
│ 1   │ 1 │ 2 │ 3 │
│ 2   │ 4 │ 5 │ 6 │

julia> CSV.read(IOBuffer("1,2,3\n4,5,6"); header=["a", "b", "c"])
2×3 DataFrames.DataFrame
│ Row │ a │ b │ c │
├─────┼───┼───┼───┤
│ 1   │ 1 │ 2 │ 3 │
│ 2   │ 4 │ 5 │ 6 │
```

An "empty" file/buffer with a header does what we'd expect:

```
julia> CSV.read(IOBuffer("a,b,c"))
0×3 DataFrames.DataFrame

```

It even has the right number of columns!

However, in the case where the CSV itself is headerless and the user supplies a header, `CSV.read` will fail on an empty buffer/file:

```julia
julia> CSV.read(IOBuffer(""), header=["a", "b", "c"])
ERROR: ArgumentError: length of provided header doesn't match number of columns of data at row 1
 in #Source#7(::Base.AbstractIOBuffer{Array{UInt8,1}}, ::CSV.Options, ::Array{String,1}, ::Int64, ::Array{DataType,1}, ::Bool, ::Bool, ::Int64, ::Int64, ::Int64, ::Bool, ::Type{T}) at /Users/gem/.julia/v0.5/CSV/src/Source.jl:120
 in (::Core.#kw#Type)(::Array{Any,1}, ::Type{CSV.Source}) at ./<missing>:0
 in #Source#6(::UInt8, ::UInt8, ::UInt8, ::String, ::Array{String,1}, ::Int64, ::Array{DataType,1}, ::Bool, ::Bool, ::Base.Dates.DateFormat, ::Int64, ::Int64, ::Int64, ::Bool, ::Type{T}, ::Base.AbstractIOBuffer{Array{UInt8,1}}) at /Users/gem/.julia/v0.5/CSV/src/Source.jl:25
 in (::Core.#kw#Type)(::Array{Any,1}, ::Type{CSV.Source}, ::Base.AbstractIOBuffer{Array{UInt8,1}}) at ./<missing>:0
 in #read#25(::Bool, ::Dict{Int64,Function}, ::Array{Any,1}, ::Function, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Type{DataFrames.DataFrame}) at /Users/gem/.julia/v0.5/CSV/src/Source.jl:298
 in (::CSV.#kw##read)(::Array{Any,1}, ::CSV.#read, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Type{DataFrames.DataFrame}) at ./<missing>:0 (repeats 2 times)
```

I'm looking into a fix for this myself, but opening an issue seems like a good first step.

I'm running Julia 0.5, in case it's relevant.

spurll:
Fixes an issue where CSV.read would fail when given an empty CSV with the header specified in a kwarg. Now returns a DataFrame with zero rows and the appropriate header.

Closes #69

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -78,7 +78,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         unsafe_read(source, UInt8) == 0xbf || seekstart(source)
     end
     datarow = datarow == -1 ? (isa(header, Vector) ? 0 : last(header)) + 1 : datarow # by default, data starts on line after header
-    rows = max(-1, rows - datarow + 1 - footerskip) # rows now equals the actual number of rows in the dataset
+    rows = fs == 0 ? 0 : max(-1, rows - datarow + 1 - footerskip) # rows now equals the actual number of rows in the dataset
 
     # figure out # of columns and header, either an Integer, Range, or Vector{String}
     # also ensure that `f` is positioned at the start of data
@@ -109,6 +109,10 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         end
         datarow != last(header)+1 && CSV.skipto!(source,last(header)+1,datarow,options.quotechar,options.escapechar)
         datapos = position(source)
+    elseif fs == 0
+        datapos = position(source)
+        columnnames = header
+        cols = length(columnnames)
     else
         CSV.skipto!(source,1,datarow,options.quotechar,options.escapechar)
         datapos = position(source)
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -56,8 +56,8 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         fullpath = "<IOBuffer>"
         fs = nb_available(source)
     elseif isa(fullpath, IO)
-        source = IOBuffer(Base.read(fullpath))
         fs = nb_available(fullpath)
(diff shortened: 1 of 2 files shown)


spurll:
Hold the phone! Found a bug in my changes! Changing this to a WIP while I track down the source.

spurll:
I found a corner-case in which the file-size is being checked and stored *after* reading from the IO stream (which implicitly assumes that the stream is rereadable). This means that non-rereadable IO gets treated as empty (which might not have previously been an issue, but can definitely interact badly with my PR).

spurll:
This should be ready to merge.

quinnj:
Thanks!

PR #72, opened 2017-03-28
loosen Compat requirement to 0.9.5

tkelman:
`take!` was introduced in 0.9.5, `transcode` and `.=` in 0.10.0 (https://github.com/JuliaLang/Compat.jl/compare/v0.9.5...v0.10.0)
but neither of those are used here


quinnj:
Thanks!

PR #75, opened 2017-04-04
Fix error on empty CSV with header and transforms

### LINKED ISSUE #74: `CSV.read` on empty buffer fails when `transforms` is specified
This is specifically an issue when the `transforms` dictionary includes functions that don't have known return values for `String` inputs. (`DataStreams.transform` attempts to infer a return type for the function and gets back `Union{}`.)

For example, say we want to `floor` the values in the second column:

```julia
julia> CSV.read(IOBuffer("a,b,c\n1.1,2.2,3.3\n4.4,5.5,6.6"); nullable=false, transforms=Dict{Int, Function}(2 => floor))
2×3 DataFrames.DataFrame
│ Row │ a   │ b   │ c   │
├─────┼─────┼─────┼─────┤
│ 1   │ 1.1 │ 2.0 │ 3.3 │
│ 2   │ 4.4 │ 5.0 │ 6.6 │
```

When there are zero rows, we get:

```julia
julia> CSV.read(IOBuffer("a,b,c"); nullable=false, transforms=Dict{Int, Function}(2 => floor))
ERROR: MethodError: Cannot `convert` an object of type Type{Union{}} to an object of type DataType
This may have arisen from a call to the constructor DataType(...),
since type constructors fall back to convert methods.
 in transform(::DataStreams.Data.Schema{true}, ::Dict{Int64,Function}) at /Users/gem/.julia/v0.5/DataStreams/src/DataStreams.jl:77
 in #stream!#5(::Array{Any,1}, ::Function, ::CSV.Source, ::Type{DataFrames.DataFrame}, ::Bool, ::Dict{Int64,Function}) at /Users/gem/.julia/v0.5/DataStreams/src/DataStreams.jl:149
 in #read#30(::Bool, ::Dict{Int64,Function}, ::Array{Any,1}, ::Function, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Type{DataFrames.DataFrame}) at /Users/gem/.julia/v0.5/CSV/src/Source.jl:302
 in (::CSV.#kw##read)(::Array{Any,1}, ::CSV.#read, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Type{DataFrames.DataFrame}) at ./<missing>:0 (repeats 2 times)
```

Note that this occurs when there is a header row but no data (or when the buffer is empty but a header is specified).

Currently the user can work around this problem by using a duck-typed wrapper function:

```julia
julia> CSV.read(IOBuffer("a,b,c"); nullable=false, transforms=Dict{Int, Function}(2 => x->floor(x)))
0×3 DataFrames.DataFrame
```

...but that seems like a bit of a kludge.

One method of solving this would be to have `CSV.read` ignore the `transforms` dictionary when the buffer is empty (I'm prepping a PR that will do just that; just needs tests). Another approach, I suppose, would be to make `DataStreams.transform` a little more resistant to to having `Union{}` pop up.

spurll:
This was specifically an issue when the transforms dictionary included
functions that didn't have known return values for `String` inputs and
when there was a header row but no data (or when the buffer was empty
but a header was specified). `DataStreams.transform` attempted to infer
a return type for the function and got back `Union{}`, which errored.

Closes #74

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -292,6 +292,10 @@ function read end
 
 function read(fullpath::Union{AbstractString,IO}, sink=DataFrame, args...; append::Bool=false, transforms::Dict=Dict{Int,Function}(), kwargs...)
     source = Source(fullpath; kwargs...)
+    if source.schema.rows == 0
+        # If the source is empty, ignore transforms to prevent type conversion errors.
+        transforms = Dict{Int,Function}()
+    end
     sink = Data.stream!(source, sink, append, transforms, args...)
     Data.close!(sink)
     return sink
@@ -299,6 +303,10 @@ end
 
 function read{T}(fullpath::Union{AbstractString,IO}, sink::T; append::Bool=false, transforms::Dict=Dict{Int,Function}(), kwargs...)
     source = Source(fullpath; kwargs...)
+    if source.schema.rows == 0
+        # If the source is empty, ignore transforms to prevent type conversion errors.
+        transforms = Dict{Int,Function}()
+    end
     sink = Data.stream!(source, sink, append, transforms)
     Data.close!(sink)
     return sink
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -290,7 +290,7 @@ sq1 = CSV.read(source, SQLite.Sink, db, "sqlite_table")
 """
 function read end
 
-function read(fullpath::Union{AbstractString,IO}, sink=DataFrame, args...; append::Bool=false, transforms::Dict=Dict{Int,Function}(), kwargs...)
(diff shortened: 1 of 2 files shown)


spurll:
As per https://codecov.io/gh/JuliaData/CSV.jl/pull/75/diff, coverage of this change is now at 100%. Should be ready to merge.

Incidentally, the change to the `CSV.read` function signature made in 2c46e57 was to address [this strange dispatch issue](https://github.com/JuliaLang/julia/issues/21481), which meant that calling `CSV.read(IOBuffer("a,b,c"))` would hit the `CSV.read{T}` method instead of the expected method.

quinnj:
Sorry I forgot about this! Thanks!

PR #78, opened 2017-04-21
Fix bug causing `CSV.read` to incorrectly see some streams as empty

### LINKED ISSUE #77: `CSV.read` sometimes incorrectly assumes IO streams are empty
In `src/Source.jl`, we can see that when `fullpath` is an `IO` (but not an `IOBuffer`), 

```julia
fs = nb_available(fullpath)
source = IOBuffer(Base.read(fullpath))
```
Prior to a recent change (in a PR that I made), this section looked like this instead:

```julia
source = IOBuffer(Base.read(fullpath))
fs = nb_available(fullpath)
```

This version suffered from the problem that, since `fullpath` was being read *before* `nb_available` was checked, `nb_available` would *always* return 0, so I flipped the lines as part of my PR to address this issue.

Unfortunately, both the original and fixed versions suffer from the issue that `nb_available` is not a good way of checking whether a *stream* (rather than an `IOBuffer`) is empty; `nb_available` just checks the number of currently buffered bytes available to be read. I've run into an issue where, when buffering is slow, there may be zero bytes buffered (so `nb_available` returns 0), but the stream isn't done, and actually reading it returns additional data. (Unfortunately, reading is skipped because CSV.jl believes that the stream is empty.)

The problem can be demonstrated (albeit clumsily) with this code:

```julia
type MultiStream{S<:IO} <: IO
    streams::Array{S}
    index::Int
end

function MultiStream{S<:IO}(streams::AbstractArray{S})
    MultiStream(streams, 1)
end

function refill(s::MultiStream)
    while eof(s.streams[s.index]) && s.index < length(s.streams)
        close(s.streams[s.index])
        s.index += 1
    end
end

function Base.close(s::MultiStream)
    for i in s.index:length(s.streams)
        close(s.streams[i])
    end
    s.index = length(s.streams)
    nothing
end

function Base.eof(s::MultiStream)
    eof(s.streams[s.index]) && s.index == length(s.streams)
end

function Base.read(s::MultiStream, ::Type{UInt8})
    refill(s)
    read(s.streams[s.index], UInt8)::UInt8
end

function Base.nb_available(s::MultiStream)
    nb_available(s.streams[s.index])
end
```

Running the current version of the code, we get this:

```julia
julia> x = MultiStream([IOBuffer(""), IOBuffer("a,b,c\n1,2,3\n"), IOBuffer(""), IOBuffer("4,5,6")]);

julia> CSV.read(x)
0×3 DataFrames.DataFrame

```

But! Since CSV.jl is *already* reading all `IO`s into `IOBuffer`s anyway, this is a very simple fix.

The problem is solved by changing this section to read:

```julia
source = IOBuffer(Base.read(fullpath))
fs = nb_available(source)
```

With this change, we get the expected result:

```julia
julia> x = MultiStream([IOBuffer(""), IOBuffer("a,b,c\n1,2,3\n"), IOBuffer(""), IOBuffer("4,5,6")]);

julia> CSV.read(x)
2×3 DataFrames.DataFrame
│ Row │ a │ b │ c │
├─────┼───┼───┼───┤
│ 1   │ 1 │ 2 │ 3 │
│ 2   │ 4 │ 5 │ 6 │
```

I plan to issue a PR along these lines in the morning.

spurll:
Previous versions assumed that nb_available could accurately check for an empty CSV, but
this doesn't work reliably for streams because nb_available only checks buffered bytes.

Closes #77.

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -56,8 +56,8 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         fullpath = "<IOBuffer>"
         fs = nb_available(source)
     elseif isa(fullpath, IO)
-        fs = nb_available(fullpath)
         source = IOBuffer(Base.read(fullpath))
+        fs = nb_available(source)
         fullpath = isdefined(fullpath, :name) ? fullpath.name : "__IO__"
     else
         source = open(fullpath, "r") do f
--- a/test/multistream.jl
+++ b/test/multistream.jl
@@ -1,0 +1,48 @@
+# Previous versions assumed that nb_available could accurately check for an empty CSV, but
+# this doesn't work reliably for streams because nb_available only checks buffered bytes
+# (see issue #77). This test verifies that even when nb_available would return 0 on a stream
+# the full stream is still read.
+
+type MultiStream{S<:IO} <: IO
+    streams::Array{S}
+    index::Int
+end
+
+function MultiStream{S<:IO}(streams::AbstractArray{S})
+    MultiStream(streams, 1)
+end
+
+function refill(s::MultiStream)
+    while eof(s.streams[s.index]) && s.index < length(s.streams)
+        close(s.streams[s.index])
+        s.index += 1
+    end
+end
+
+function Base.close(s::MultiStream)
+    for i in s.index:length(s.streams)
+        close(s.streams[i])
+    end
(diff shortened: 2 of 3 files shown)


omus:
LGTM

quinnj:
This is great; thanks!

PR #89, opened 2017-08-07
Fix invalid escape sequence in docstring, update CI for 0.6

cjprybol:
CSV won't load for me on 0.7-DEV. Adding the extra backslash in the docstring fixes the parsing. 

Not sure if the CI file updates are correct, but I presume `nightly` is now testing against 0.7-DEV rather than 0.6 so this is my best guess for how to add 0.6 CI support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ os:
 
 julia:
   - 0.5
+  - 0.6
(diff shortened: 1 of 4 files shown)


quinnj:
Thanks!

PR #94, opened 2017-08-26
Update REQUIRE with WeakRefStrings & DataStreams upperbounds for release

cjprybol:


cjprybol:
holding off on this until https://github.com/JuliaLang/METADATA.jl/pull/10915. If the upper bounds are added as shown in [this diff](https://github.com/JuliaLang/METADATA.jl/pull/10915/files#diff-ab8d73f7d0899155c184443481c0c718) then this won't be necessary. In that case, remember to close https://github.com/JuliaLang/METADATA.jl/pull/10923 too rather than retagging a v0.1.5 release

nalimilan:
Adding the bounds here to match those in METADATA.jl would be a good idea. We can easily remove them once the package will support the newer versions.

PR #102, opened 2017-09-14
Fix #43 by supporting CategoricalArray parsing.

### LINKED ISSUE #43: Support CategoricalArray parsing

quinnj:
cc @nalimilan 

This ended up being more minimal than I expected. I know we're cheating a bit by just returning a WeakRefString from the `CSV.parsefield(io, ::Type{CategoricalValue})` function, but in the DataStreams system, I think this should be fine. The important part is the Source's Data.schema have `CategoricalValue` in the types and that the Sink be able to handle them. I still need to do some more Sink testing, but so far, they're already setup to handle this just fine.

Fixes #43.

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -14,6 +14,7 @@ function Source(fullpath::Union{AbstractString,IO};
               decimal=PERIOD,
               truestring="true",
               falsestring="false",
+              categorical::Bool=true,
 
               footerskip::Int=0,
               rows_for_type_detect::Int=20,
@@ -28,7 +29,7 @@ function Source(fullpath::Union{AbstractString,IO};
                                             quotechar=typeof(quotechar) <: String ? UInt8(first(quotechar)) : (quotechar % UInt8),
                                             escapechar=typeof(escapechar) <: String ? UInt8(first(escapechar)) : (escapechar % UInt8),
                                             null=null, dateformat=dateformat, decimal=decimal, truestring=truestring, falsestring=falsestring),
-                        header=header, datarow=datarow, types=types, nullable=nullable, footerskip=footerskip,
+                        header=header, datarow=datarow, types=types, nullable=nullable, categorical=categorical, footerskip=footerskip,
                         rows_for_type_detect=rows_for_type_detect, rows=rows, use_mmap=use_mmap)
 end
 
@@ -39,6 +40,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
                 datarow::Int=-1, # by default, data starts immediately after header or start of file
                 types=Type[],
                 nullable::Union{Bool, Null}=null,
+                categorical::Bool=true,
 
                 footerskip::Int=0,
                 rows_for_type_detect::Int=20,
@@ -131,6 +133,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         columntypes = types
     elseif isa(types, Dict) || isempty(types)
         columntypes = Vector{Type}(cols)
+        categoricals = Dict{Int, Set{String}}(i=>Set{String}() for i = 1:cols)
         fill!(columntypes, Any)
         lineschecked = 0
         while !eof(source) && lineschecked < min(rows < 0 ? rows_for_type_detect : rows, rows_for_type_detect)
@@ -138,7 +141,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
             # println("type detecting on row = $lineschecked...")
             for i = 1:cols
                 # print("\tdetecting col = $i...")
-                typ = CSV.detecttype(source, options, columntypes[i])::Type
+                typ = CSV.detecttype(source, options, columntypes[i], categoricals[i])::Type
                 # print(typ)
                 columntypes[i] = CSV.promote_type2(columntypes[i], typ)
                 # println("...promoting to: ", columntypes[i])
@@ -149,6 +152,13 @@ function Source(;fullpath::Union{AbstractString,IO}="",
             options = Options(delim=options.delim, quotechar=options.quotechar, escapechar=options.escapechar,
                               null=options.null, dateformat=Dates.ISODateTimeFormat, decimal=options.decimal,
                               datarow=options.datarow, rows=options.rows, header=options.header, types=options.types)
+        end
+        if categorical
+            for i = 1:cols
+                T = columntypes[i]
+                columntypes[i] = ifelse(length(categoricals[i]) / rows_for_type_detect < .9 && T <: WeakRefString,
+                    CategoricalValue{String, UInt16}, T)
+            end
         end
     else
         throw(ArgumentError("$cols number of columns detected; `types` argument has $(length(types)) entries"))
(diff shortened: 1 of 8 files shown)


src/Source.jl
```diff
@@ -150,6 +153,13 @@ function Source(;fullpath::Union{AbstractString,IO}="",
                               null=options.null, dateformat=Dates.ISODateTimeFormat, decimal=options.decimal,
                               datarow=options.datarow, rows=options.rows, header=options.header, types=options.types)
         end
+        if categorical
+            for i = 1:cols
+                T = columntypes[i]
+                columntypes[i] = ifelse(length(categoricals[i]) / rows_for_type_detect < .9 && T <: WeakRefString,
+                    CategoricalValue{String, UInt16}, T)
```
nalimilan: I suggested using `UInt16` only if there was a mechanism to start again with a larger type in case of overflow. As long as that mechanism isn't implemented, it would be safer to use `UInt32` even if that's wasteful most of the time.

Also, incorrect indentation?

src/Source.jl
```diff
@@ -230,6 +240,7 @@ Keyword Arguments:
 * `append::Bool=false`: if the `sink` argument provided is an existing table, `append=true` will append the source's data to the existing data instead of doing a full replace
 * `transforms::Dict{Union{String,Int},Function}`: a Dict of transforms to apply to values as they are parsed. Note that a column can be specified by either number or column name.
 * `transpose::Bool=false`: when reading the underlying csv data, rows should be treated as columns and columns as rows, thus the resulting dataset will be the "transpose" of the actual csv data.
+* `categorical::Bool=true`: read string columns as CategoricalArrays, as long as the % of unique values seen during type detection isn't 100%
```
nalimilan: Need backticks around `CategoricalArray` (a link could be useful too). Also shouldn't 100% be "below 90%"? BTW, maybe that threshold is a bit high? 75% could be more reasonable? Finally, it would be nice to explain that this will reduce memory use dramatically when the number of unique values is small.

src/io.jl
```diff
@@ -215,7 +215,7 @@ promote_type2(::Type{Null}, ::Type{WeakRefString{UInt8}}) = Union{WeakRefString{
 promote_type2(::Type{Any}, ::Type{Null}) = Null
 promote_type2(::Type{Null}, ::Type{Null}) = Null
 
-function detecttype(io, opt::CSV.Options{D}, prevT) where {D}
+function detecttype(io, opt::CSV.Options{D}, prevT, seen) where {D}
```
nalimilan: I'd rather call this `uniques` or `levels` (`seen` is used in `Base.unique` for a set indicating whether a value was seen). Same for `categoricals` above (using the same name for both would be clearer).

src/Source.jl
```diff
@@ -131,14 +133,15 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         columntypes = types
     elseif isa(types, Dict) || isempty(types)
         columntypes = Vector{Type}(cols)
+        categoricals = Dict{Int, Set{String}}(i=>Set{String}() for i = 1:cols)
```
nalimilan: Wouldn't using a `WeakRefString` be OK here to avoid some allocations?

src/io.jl
quinnj: updated

src/Source.jl
quinnj: updated

src/Source.jl
quinnj: fair enough

src/Source.jl
quinnj: done

src/Source.jl
```diff
@@ -230,6 +240,7 @@ Keyword Arguments:
 * `append::Bool=false`: if the `sink` argument provided is an existing table, `append=true` will append the source's data to the existing data instead of doing a full replace
 * `transforms::Dict{Union{String,Int},Function}`: a Dict of transforms to apply to values as they are parsed. Note that a column can be specified by either number or column name.
 * `transpose::Bool=false`: when reading the underlying csv data, rows should be treated as columns and columns as rows, thus the resulting dataset will be the "transpose" of the actual csv data.
+* `categorical::Bool=true`: read string column as a `CategoricalArray` ([ref](https://github.com/JuliaData/CategoricalArrays.jl)), as long as the % of unique values seen during type detection is less than 67%. This will dramatically reduce memory use in cases where the number of unique values is small.
```
nalimilan: Why not put the link directly on `CategoricalArray`? Also, plural "columns" and replace "the %" with "the proportion"?

alyst:
I don't have a strong opinion on `categorical=true`, but for my own applications I would rather set it to `false`.

I understand the intention is to improve the speed and memory usage for the most common use cases.
The downside of guessing is that predictions are not very stable, even for the tables of the same format (so the script that was working fine before may suddenly fail on the new dataset and cause a lot of confusion for the inexperienced user).
JuliaData/CategoricalArrays.jl#77 is trying to solve this by making `CategoricalValue{T} <: AbstractString` at the expense of constraining `T<:AbstractString` in the near future.

What about the alternative solution: `categorical=false`, but also `diagnose=:auto/true/false` option is introduced.
If `diagnose=true`, after the table is loaded, its columns would be checked for any optimization possibilities and if there's anything to improve, it would report something like
```
Table took 15s to load, occupies 120Mb in memory.
Consider using categorical=true or declaring the following columns as categorical to improve loading times and memory usage:
:A (String): 4 unique values (80% duplicates)
....
``` 
If `diagnose=:auto` (the default), the diagnostic is triggered only if the table took too long to load or occupies a lot of memory (probably relative to its size).

So in the default scenario for the small tables the user would see nothing.
For the big tables the user would see the notification and either set `categories=true`, fine tune `types=` and/or just disable annoying notification with `diagnose=false`.

That way there's no guessing by default, but any performance issues are revealed, and the fix is explicitly provided by the user.

nalimilan:
I suggest we keep `categorical=true` for some time and see how it goes. In theory, changing the type depending on the input values is potentially problematic, but in practice I wonder whether the threshold we have retained (67% of unique values in the 100 first rows) should give relatively stable guesses. It may not be common to have data which is just around that threshold so that it goes above or below depending on specific data sets. Do you think that's the case in your data? We could also tweak that threshold to make it more conservative.

Printing diagnostics which depend on the values by default would expose users to getting verbose messages only in some cases, so that's not really better than changing the type.

alyst:
>  It may not be common to have data which is just around that threshold so that it goes above or below depending on specific data sets. Do you think that's the case in your data?

I think the "shaky" case might be actually quite common. If one converts M x N table into a "long" format (i.e. a table with M * N rows corresponding to each cell of the original table), then the column that contains IDs of the rows of the original M x N table will have all unique values in the first M rows. So if `nguess` < (1+o)*M, the column would be a string, if `nguess` > ~1.5 * M (or something like that), it would be a category. If empty cells are removed during "wide" -> "long" conversion, it becomes even less predictable.
But such tables could be quite large. To guess it right you would rather have to set the default threshold to be a fraction of the total number of rows (1-5%).   
The actual examples are `evidence.txt` tables generated by [MaxQuant](http://www.coxdocs.org/doku.php?id=maxquant:start) (essentially it's experiment (10-1000) X peptide (~10^4) table in the "long" format).

That's also the case where `diagnose` approach would get it right since the diagnostic would be based on the complete data.

> Printing diagnostics which depend on the values by default would expose users to getting verbose messages only in some cases, so that's not really better than changing the type.

The key difference is that the user would have to make his/her own decision regarding the column types, and that decision will persist for all subsequent input data.
Even if the decision would be `categorical=true` or `diagnose=false`, it would be explicitly stated in the script (as a notice to the author and any other users of the script).

`diagnose` would also provide continuous monitoring of manual column types specifications, which is useful e.g. to detect changes in the input data format.

nalimilan:
Yes, in general with sorted data the proportion of unique values isn't a great indicator if the number of rows used to guess the type is small compared to the number of levels. The heuristic more or less relies on the assumption that the first rows are vaguely representative of the full data set. Without that assumption I don't see how we could predict anything about the kind of data a column contains. Unfortunately AFAIK reading random rows in the file would be slow.

PR #110, opened 2017-10-04
Fixed #103 and added test for same

### LINKED ISSUE #103: options has no field datecheck
[This line](https://github.com/JuliaData/CSV.jl/blob/master/src/Sink.jl#L56) tries to access the `datecheck` field of the `Options` struct, but this field does not exist.

ExpandingMan:
The test is a little wonky but at least now we'll see if any critical errors get thrown when writing dataframes (which seems extremely important).

I *think* the errors are from using an old version of DataFrames and/or DataStreams.


nalimilan:
It looks like @quinnj removed that field on purpose when porting to new DataStreams, so the minimalist approach taken by https://github.com/JuliaData/CSV.jl/issues/129 might be more appropriate.

quinnj:
Fixed on master w/ tests.

PR #115, opened 2017-11-26
Type detection fixes

alyst:
Fixes a few issues with type detection
* use `CategoricalArrays.catvaluetype()` to return the value of categorical (should be `CategoricalString` for `WeakRefString`, previously was `CategoricalValue`)
* start counting column levels from the very beginning, otherwise when the column is detected to be `WeakRefString` many levels are skipped
* allow categorical type for the columns with missing values
* ignore rows with missing values when calculating the fraction of unique values (for that `levels::Set{Str}` changed into `levels::Dict{Str, Int}` (the last 3 bullet points tested in `baseball.csv`)
* fix `promote_rule2(Union{T, Missing}, WeakRefString)`, should be `Union{WeakRefString, Missing}` (was `WeakRefString`) -- that's where `attenu.csv` of RDatasets was failing

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -175,15 +175,15 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         end
     end
     if !ismissing(nullable)
-        if nullable
+        if nullable # allow missing values in all columns
             for i = 1:cols
                 T = columntypes[i]
-                columntypes[i] = ifelse(T >: Missing, T, Union{T, Missing})
+                columntypes[i] = Union{Missings.T(T), Missing}
             end
-        else
+        else # disallow missing values in all columns
             for i = 1:cols
                 T = columntypes[i]
-                columntypes[i] = ifelse(T >: Missing, Missings.T(T), T)
+                columntypes[i] = Missings.T(T)
             end
         end
     end
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -157,8 +157,13 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         if categorical
             for i = 1:cols
                 T = columntypes[i]
-                columntypes[i] = ifelse(length(levels[i]) / rows_for_type_detect < .67 && T <: WeakRefString,
-                    CategoricalValue{String, UInt32}, T)
+                if length(levels[i]) / rows_for_type_detect < .67 &&
+                        T !== Missing && Missings.T(T) <: WeakRefString
+                    columntypes[i] = CategoricalArrays.catvaluetype(Missings.T(T), UInt32)
+                    if T >: Missing
+                        columntypes[i] = Union{columntypes[i], Missing}
+                    end
+                end
             end
         end
     else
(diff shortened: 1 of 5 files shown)


alyst:
CI failures look unrelated: AppVeyor win32 builds because the inferred column types are `Int32`, while the test expects `Int64` and nightlies due to `Date`

test/source.jl
```diff
@@ -439,3 +439,8 @@ ds = CSV.read(f)
 # #107
 df = CSV.read(IOBuffer("1,a,i\n2,b,ii\n3,c,iii"); datarow=1)
 @test size(df) == (3, 3)
+
+# #115
+df = CSV.read(joinpath(dir, "attenu.csv"), null="NA", rows_for_type_detect=200)
```
nalimilan: BTW, `null` should really be renamed.

alyst: WRT `null`, there's also a number of `nullable` occurrences. What's the current recommended term (adjective) for a fact that a container does/can store missing values?
For example, `MissingCategoricalArray` is a little bit misleading in that regard.

nalimilan: I've not been able to come up with a good and short replacement for `nullable` so far... :-/

alyst: [missable](http://www.yourdictionary.com/missable)?

src/Source.jl
```diff
@@ -235,7 +239,7 @@ Keyword Arguments:
 * `types`: column types can be provided manually as a complete Vector{Type}, or in a Dict to reference individual columns by name or number
 * `nullable::Bool`: indicates whether values can be nullable or not; `true` by default. If set to `false` and missing values are encountered, a `Data.NullException` will be thrown
 * `footerskip::Int`: indicates the number of rows to skip at the end of the file
-* `rows_for_type_detect::Int=100`: indicates how many rows should be read to infer the types of columns
+* `rows_for_type_detect::Int=20`: indicates how many rows should be read to infer the types of columns
```
nalimilan: I hadn't noticed this. I realize this is just a fix to match the actual behavior, but I think we should raise this value. Something like 100 or 200 sounds safer, and shouldn't impact performance a lot, should it @quinnj?

alyst: IIUC most of the time is spent compiling methods for the new scheme.
I've tried optimizing the `rows_for_type_detect` for johnmyleswhite/RDatasets.jl#48, but that didn't really improve the test time (~5 min for 733 datasets).

test/source.jl
nalimilan: "Able to avoid detection."? :-)

I could find [one](https://rdrr.io/cran/rsolr/src/R/SolrExpression-class.R) precedent in relation with missing values, but that's the only one...

tlnagy:
Bump @nalimilan. It would be great to have this merged so that `RDatasets` can be updated.

tlnagy:
Hey @quinnj, did you get a chance to look at this?

quinnj:
Thanks! This is great!

alyst:
@quinnj Thanks for merging! When do you plan to tag the new release?

PR #124, opened 2017-11-29
Replace all uses of null with missing

nalimilan:
In particular, rename the `null` argument to `missingstring` for consistency with
`truestring` and `falsestring`. This is breaking.

It would have been better to break things in 0.2.0, but better do it sooner than later.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -77,14 +77,14 @@ using CSV, TextParse
 for T in (Int, Float64, WeakRefStrings.WeakRefString{UInt8}, Date, DateTime)
     println("comparing for T = $T...")
     # T == WeakRefStrings.WeakRefString{UInt8} && continue
-    @time CSV.read("/Users/jacobquinn/Downloads/randoms_$(T).csv"; nullable=true);
+    @time CSV.read("/Users/jacobquinn/Downloads/randoms_$(T).csv"; allowmissing=true);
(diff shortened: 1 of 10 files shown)


quinnj:
I'd rather not do this; `null` is a pretty good name for what this is, even if julia itself uses `missing`

nalimilan:
It would be really weird to talk about `null` in the context of CSV, where it isn't a common term (R saves missing values as "NA" for example), but use `missing` everywhere else. How are people supposed to know what "null" means if we don't use it anywhere else?

alyst:
`allowmissing` nee `nullable` is subtle:
* docstring doesn't reflect that its default value is now `missing` meaning "don't touch autodetected/user specified column types", i.e. **allow** missing
* `true` **forces** rather than allows all columns to be of type `Union{T, Missing}`

Maybe change it to be `::Symbol = :allow/:force/:no`? Rename to `missingmode` (:allow/:force/:error)?

nalimilan:
@alyst Agreed, that was the second issue I wanted to raise. Could also be something like `:yes`/`:auto`/`:no`.

alyst:
@nalimilan Ah, if there's `allowmissing!()` method, then `allowmissing=` should stay, of course. Maybe with `:all/:auto/:no` options.

nalimilan:
Done by https://github.com/JuliaData/CSV.jl/pull/173.

PR #125, opened 2017-12-01
Fixed quoting issues

### LINKED ISSUE #119: Quotes around strings and vectors
When I save a `DataFrame` with strings in cells or vectors in cells, `CSV.write` doesn't put `"` quotes around the strings or vectors like `DataFrames.writetable` does. Ex:

```
df = DataFrame(A = 1:3, B = [4,6.0,2], C = ["a","b","c"], D = [[4,5],[3,3],[4,4]])
CSV.write("x.csv", df)
```

`x.csv` doesn't have quotes around the cells
```
A,B,C,D
1,4.0,a,[4, 5]
2,6.0,b,[3, 3]
3,2.0,c,[4, 4]
```

And it gets read in wrong when you it back `CSV.read("x.csv")`, because of the commas inside the square brackets.

dawbarton:
Fixes #119 

Converts UInt8 fields into Char at the point of use (it was partially done before but not uniformally) and removes an unnecessary specialism of `Data.streamto!` to enable automatic (or mandatory) quoting for general data structures (e.g., arrays).

### DIFF
--- a/src/Sink.jl
+++ b/src/Sink.jl
@@ -21,7 +21,7 @@ quoted(::Type{Val{false}}, val, q, e, d) = (q in val || d in val) ? string(q, re
 
 function writeheaders(io::IOBuffer, h::Vector{String}, options, quotefields)
     cols = length(h)
-    q = options.quotechar; e = options.escapechar; d = options.delim
+    q = Char(options.quotechar); e = Char(options.escapechar); d = Char(options.delim)
     for col = 1:cols
         Base.write(io, quoted(quotefields, h[col], q, e, d), ifelse(col == cols, NEWLINE, options.delim))
     end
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks. Would it make sense to change the `quotechar` field to be a `Char` in the first place?

dawbarton:
Yes, that can be easily done. The only reason I didn't do that was because I assumed there must have been a reason for using `UInt8` rather than a `Char` in the first place and I didn't want to change it without knowing why.

If there is no reason (other than historical) then I can easily put together a PR that makes the appropriate changes.

quinnj:
All "character" fields in CSV have been UInt8 for consistency in parsing/writing scenarios, so please don't change that. I'm tied up for about a week, but I'm planning on diving into CSV as soon as I can to review this and other issues that have been filed.

dawbarton:
No problem. In which case the fix I've proposed is (hopefully!) the appropriate one.

nalimilan:
So that means non-ASCII characters are not supported, right?

quinnj:
let's have CI re-run on this, but otherwise looks good to me.

rand5:
With this fix I am getting the desired functionality on my PC but on a mac the quote character is still being ignored

vvjn:
Why is this still not merged? It doesn't seem to be that complicated a change. I literally have a copy of `DataFrames.readtable` in my own codebase because issue #119 is still not fixed.

PR #138, opened 2017-12-20
Deduplicate column types detection code

alyst:
The column type detection code in `Source` and `TransposedSource` is identical (except file position updates), so I put it into `detect_dataschema()` function to reduce the overhead of keeping them in sync.

The CI might be broken, because it's also broken in master. I will update/fix the PR when the master is back to normal.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -3,7 +3,7 @@ module CSV
 
 using DataStreams, WeakRefStrings, Missings, CategoricalArrays, DataFrames
 
-using Compat, Compat.Mmap, Compat.Dates
+using Compat, Compat.Mmap, Compat.Dates, Compat.equalto
 
 struct ParsingException <: Exception
     msg::String
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -51,12 +51,12 @@ Keyword Arguments:
  * `quotechar::Union{Char,UInt8}`: the character that indicates a quoted field that may contain the `delim` or newlines; default `'"'`
  * `escapechar::Union{Char,UInt8}`: the character that escapes a `quotechar` in a quoted field; default `'\\'`
  * `missingstring::String`: indicates how missing values are represented in the dataset; default `""`
- * `dateformat::Union{AbstractString,Dates.DateFormat}`: how dates/datetimes are represented in the dataset; default `Base.Dates.ISODateTimeFormat`
+ * `dateformat::Union{Dates.DateFormat, Nothing}`: how dates/datetimes are represented in the dataset; default `Base.Dates.ISODateTimeFormat`
  * `decimal::Union{Char,UInt8}`: character to recognize as the decimal point in a float number, e.g. `3.14` or `3,14`; default `'.'`
  * `truestring`: string to represent `true::Bool` values in a csv file; default `"true"`. Note that `truestring` and `falsestring` cannot start with the same character.
  * `falsestring`: string to represent `false::Bool` values in a csv file; default `"false"`
 """
-struct Options{D}
+struct Options{D <: Union{DateFormat, Nothing}}
     delim::UInt8
     quotechar::UInt8
     escapechar::UInt8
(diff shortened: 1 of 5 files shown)


alyst:
In the master `options::CSV.Options` is allowed to be changed during column type detection. Specifically, `options.dateformat` may change from `missing` to `ISODateTimeFormat` if there's any column of `Dates.TimeType` type. The problem is that if the other columns (including the user specified ones) are of type `Date`, they would also be parsed as `DateTime`, although the user have never specified that.

In this PR the behaviour is changed:
 * once constructed, `options` are never modified
 * the default value of `options.dateformat` is changed to `nothing`
 * if `options.dateformat === nothing`, each date/datetime column is being parsed according to its individual type
 * `D` type parameter of `Options` is removed as it should not have impact on the performance

quinnj:
> D type parameter of Options is removed as it should not have impact on the performance

I'm not sure this is right. I specifically added the type parameter because otherwise when you try to parse Date/DateTime, there's a type instability when referencing `options.dateformat` unless it's included in the `Options` type parameter.

alyst:
> I specifically added the type parameter because otherwise when you try to parse Date/DateTime, there's a type instability when referencing options.dateformat unless it's included in the Options type parameter.

Yes, it's true for `Options{D}`.
I've done benchmarks on 0.6: `parsefield()` code in PR is 2x slower (2x more allocations, 1.5x more memory used).
But that's not too bad, I feared 10x slowdown.
On my 0.7 there's some trouble with dates parsing, everything is much slower, so the difference is just 10%.

Definitely this type-instability needs to be fixed, but I think the proper solution would be to make it column-specific.
(The idea is to have `columntypes::Vector{<:AbstractColumnData}` with the default `ColumnData{T} <: AbstractColumnData` and `DateTimeData{D <: DateFormat} <: AbstractColumnData`.
`parsefield()` would take `AbstractColumnData` argument).

ATM the user can specify `dateformat=`, which would be used for all datetime columns, and that's ok (although it would be nice to allow to specify it per column).
What's not ok is that for `dateformat=missing` (the default) `CSV.read()` uses the detected format of the first datetime column, instead of allowing any datetime format.

So my idea was to fix this logic and then implement per-column formats in a separate PR.
I can also revert `{D}` removal for the `Options` to fix `parsefield()` instability, but I'm not sure `D` should be kept for `Source` and `Sink`.

alyst:
I've rebased the PR onto the current master taking all type detection code updates. Some extra tiny fixes are in additional commits.
I reverted the removal of `D` type parameter from `Options`, just added some type constraints (which allowed to identify `TransformedSource(..., dateformat=missing)` issue).
`D` is still removed from `Source` and `Sink`, but it should not have an effect on `parsefield()`.

src/TransposedSource.jl
```diff
@@ -193,86 +193,13 @@ function TransposedSource(;fullpath::Union{AbstractString,IO}="",
...
-            for i = autocols
-                T = columntypes[i]
-                columntypes[i] = Missings.T(T)
-            end
-        else
-            throw(ArgumentError("allowmissing must be either :all, :none or :auto"))
-        end
-    end
+    sch, fixed_options = detect_dataschema(source, columnnames, types, options,
+                            allowmissing, categorical, weakrefstrings,
```
nalimilan: Align arguments.

src/Source.jl
```diff
@@ -141,83 +141,11 @@ function Source(;fullpath::Union{AbstractString,IO}="",
...
-            for i = autocols
-                T = columntypes[i]
-                columntypes[i] = Missings.T(T)
-            end
-        else
-            throw(ArgumentError("allowmissing must be either :all, :none or :auto"))
-        end
-    end
+    sch, fixed_options = detect_dataschema(source, columnnames, types, options,
+                            allowmissing, categorical, weakrefstrings,
```
nalimilan: Align arguments.

src/CSV.jl
```diff
@@ -3,7 +3,7 @@ module CSV
 
 using DataStreams, WeakRefStrings, Missings, CategoricalArrays, DataFrames
 
-using Compat, Compat.Mmap, Compat.Dates
+using Compat, Compat.Mmap, Compat.Dates, Compat.equalto
```
nalimilan: Use `isequal` now instead of `equalto`.

quinnj:
This is better now on master and I have a local branch that cleans things up further.

PR #139, opened 2017-12-23
Raise rows_for_type_detect default value to 100

nalimilan:
20 is really low, especially for detection of the presence of missing values.


quinnj:
This is fine, but ultimately, we should probably base this on how many rows are in the actual dataset.

PR #140, opened 2017-12-23
Revert addition of weakrefstrings option

nalimilan:
It broke tests, and has weird semantics since it completely overrides
explicit request for WeakRefString columns via the types argument.
This partly reverts commit ea2f215860dc0fec9eca09b1ea02763e683f5e80.

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -15,7 +15,6 @@ function Source(fullpath::Union{AbstractString,IO};
               truestring="true",
               falsestring="false",
               categorical::Bool=true,
-              weakrefstrings::Bool=false,
 
               footerskip::Int=0,
               rows_for_type_detect::Int=20,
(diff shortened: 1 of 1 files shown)


alyst:
+1 for reverting it.

If this is the right context to discuss the issue: I think `WeakRefString` is a proper optimization, but we should keep this type as internal as possible:
 * stop exporting it from `WeakRefStrings.jl` (`CSV.jl` should explicitly import it)
 * do not mention it in the user docs and do not add `weakrefstrings` option to `CSV`. To my understanding, with quinnj/WeakRefStrings.jl#17 (and quinnj/WeakRefStrings.jl#20) there's no easy way to leak `WeakRefString` object outside of `WeakRefStringArray`, unless indexing `.elements` directly. If I'm wrong, then we should do the transform to `String` unconditionally.
* I don't think `WeakRefString` is a type that should be allowed to be specified in `types=`: a) the returned eltype for this column would be `String`; b) anyway, if it's not exported and not advertized in the docs, it's very unlikely that somebody will use it for `types=`

There's also `DataStreams.weakrefstrings()` method that IIUC already defaults to `false` for generic sinks and to `true` for `CSV.Sink`.
In theory, it should already facilitate `WeakRefString` => `String` transform for the appropriate sinks.

quinnj:
@alyst, yes, I had some offline travel time today and basically came to the same conclusions/plan. Basically, always use WeakRefStrings for parsing/internally, but to the user, they should essentially never see an actual WeakRefString.

nalimilan:
Makes sense. Though it could still be useful to have an argument to allow the user to choose whether to get a `WeakRefStringArray` or an `Array{String}`? What are the implications of keeping `WeakRefStringArray` columns in a data frame? Do they force keeping the whole CSV file in memory? Can they be serialized?

quinnj:
Leaving the option for now; WeakRefStrings always return Strings now, so WeakRefString usage should be almost entirely an internal implementation detail for users. Leaving the keyword arg as an escape hatch in case users really need to avoid having WeakRefStrings.

PR #143, opened 2018-01-06
0.6/0.7 compat & various minor fixes

quinnj:


src/Source.jl
```diff
@@ -86,7 +84,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
     datarow = datarow == -1 ? (isa(header, Vector) ? 0 : last(header)) + 1 : datarow # by default, data starts on line after header
     rows = fs == 0 ? -1 : max(-1, rows - datarow + 1 - footerskip) # rows now equals the actual number of rows in the dataset
 
-    # figure out # of columns and header, either an Integer, Range, or Vector{String}
+    # figure out # of columns and header, either an Integer,  AbstractRange, or Vector{String}
```
alyst: extra ws

src/Source.jl
```diff
@@ -239,7 +239,7 @@ Keyword Arguments:
 * `decimal::Union{Char,UInt8}`: character to recognize as the decimal point in a float number, e.g. `3.14` or `3,14`; default `'.'`
 * `truestring`: string to represent `true::Bool` values in a csv file; default `"true"`. Note that `truestring` and `falsestring` cannot start with the same character.
 * `falsestring`: string to represent `false::Bool` values in a csv file; default `"false"`
-* `header`: column names can be provided manually as a complete Vector{String}, or as an Int/Range which indicates the row/rows that contain the column names
+* `header`: column names can be provided manually as a complete Vector{String}, or as an Int/ AbstractRange which indicates the row/rows that contain the column names
```
alyst: ws

src/TransposedSource.jl
```diff
@@ -233,6 +234,9 @@ function TransposedSource(;fullpath::Union{AbstractString,IO}="",
             columntypes[c] = typ
         end
     end
+    if !weakrefstrings
+        columntypes = [T <: WeakRefString ? String : T for T in columntypes]
```
alyst: Since `T` could be e.g. `Union{Missing, WeakRefString{UInt8}}` (and `WeakRefStringArray` now supports that) the code here should do the similar thing as a little bit above when converting to categorical.

src/Source.jl
```diff
@@ -251,10 +251,7 @@ Keyword Arguments:
...
 * `categorical::Bool=true`: read string column as a `CategoricalArray` ([ref](https://github.com/JuliaData/CategoricalArrays.jl)), as long as the % of unique values seen during type detection is less than 67%. This will dramatically reduce memory use in cases where the number of unique values is small.
-
-Note by default, "string" or text columns will be parsed as the [`WeakRefString`](https://github.com/quinnj/WeakRefStrings.jl) type. This is a custom type that only stores a pointer to the actual byte data + the number of bytes.
-To convert a `String` to a standard Julia string type, just call `string(::WeakRefString)`, this also works on an entire column.
-Oftentimes, however, it can be convenient to work with `WeakRefStrings` depending on the ultimate use, such as transfering the data directly to another system and avoiding all the intermediate copying.
+* `weakrefstrings::Bool=true`: whether WeakRefStrings should be used internally to speed up file parsing; can only be `=true` for Sinks that support WeakRefStringArrays; note that regular Strings are returned from WeakRefStringArray; WeakRefStrings are only used internally.
```
alyst: "WeakRefStringArrays" is not a proper type name, but maybe for clarity it's possible to adjust the sentence a bit, so that it uses "`WeakRefString`", "`String`" etc, e.g

```
whether to use [`WeakRefStrings`](https://github.com/quinnj/WeakRefStrings.jl) package to speed up file parsing; can only be `=true` for the `Sink` objects that support `WeakRefStringArray` columns. Note that `WeakRefStringArray` still returns regular `String` elements.
```

src/TransposedSource.jl
alyst: The same thing should be done for `Source`.
There's also #138 to consolidate column type detection code.

src/Source.jl
quinnj: addressed

src/TransposedSource.jl
quinnj: addressed

src/Source.jl
quinnj: addressed

src/Source.jl
quinnj: addressed

src/CSV.jl
```diff
@@ -38,6 +38,9 @@ end
     return byte
 end
 
+substitute(::Type{Union{T, Missing}}, ::Type{T1}) where {T, T1} = Union{T1, Missing}
```
alyst: Actually, that's the kind of function that I wanted to have in `Missings`/0.7 Base, because it's implicitly used in several places at JuliaData (cc @nalimilan).
Maybe it should be called `substitute_type`?

src/CSV.jl
```diff
@@ -38,6 +38,9 @@ end
...
+substitute(::Type{T}, ::Type{T1}) where {T, T1} = T1
```
alyst: I guess there also should be
```julia
substitute(::Type{Missing}, ::Type{T1}) where {T1} = Missing
```
rule.

src/Source.jl
```diff
@@ -179,7 +175,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         end
     end
     if !weakrefstrings
-        columntypes = [T <: WeakRefString ? String : T for T in columntypes]
+        columntypes = [Missings.T(T) <: WeakRefString ? substitute(T, String) : T for T in columntypes]
```
alyst: That's where `substitute(Missing, T)` rule is required, because `Missings.T(Missing) == Union{} <: WeakRefString`.

src/Source.jl
```diff
@@ -156,12 +156,8 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         if categorical
             for i = 1:cols
                 T = columntypes[i]
-                if length(levels[i]) / sum(values(levels[i])) < .67 &&
-                        T !== Missing && Missings.T(T) <: WeakRefString
-                    columntypes[i] = CategoricalArrays.catvaluetype(Missings.T(T), UInt32)
-                    if T >: Missing
-                        columntypes[i] = Union{columntypes[i], Missing}
-                    end
+                if length(levels[i]) / sum(values(levels[i])) < .67 && Missings.T(T) <: WeakRefString
```
alyst: `T !== Missing` check was here to exclude all-missing columns from the conversion to categorical.

src/CSV.jl
nalimilan: In Base the convention seems to be `type*`, e.g. `typeintersect` and `typejoin` (we have discussed adding `typesubtract`). So why not `typesubstitute` (but of course it's much less general than the others).

alyst: It's true, for the Base it should be more specific. `substitute_nonmissing_type`? `nonmissingtype_substitute`? Don't have a good variant that starts with `type`

nalimilan: I'm not sure we'll be able to include all of these helper functions in Base, so for now it could be just `Missings.typesubstitute` or something like that.

Issue #159, opened 2018-01-25
nullable: inconsistent docs vs behaviour; MethodError when nullable==true


mvhulten:
I have installed Julia 0.6.2, and CSV.jnl 0.2.1.

I have a CSV file with missing values:
```
$ cat test.csv 
x,y,z
1,2,3
1,,3
1,,3
1,2,
```
Then I try to read this file:
```
julia> test = CSV.read("test.csv"; datarow=2, types=[Int,Int,Int])
ERROR: MissingException: encountered a missing value for a non-null column type on row = 2, col = 2
Stacktrace:
 [1] (::CSV.##4#5)(::Int64, ::Int64) at /home/mhu027/.julia/v0.6/CSV/src/parsefields.jl:122
 [2] parsefield(::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Type{Int64}, ::CSV.Options{Void}, ::Int64, ::Int64, ::Base.RefValue{CSV.ParsingState}, ::CSV.##4#5) at /home/mhu027/.julia/v0.6/CSV/src/parsefields.jl:164
 [3] macro expansion at /home/mhu027/.julia/v0.6/DataStreams/src/DataStreams.jl:542 [inlined]
 [4] stream!(::CSV.Source{Base.AbstractIOBuffer{Array{UInt8,1}},Void}, ::Type{DataStreams.Data.Field}, ::DataFrames.DataFrameStream{Tuple{Array{Int64,1},Array{Int64,1},Array{Int64,1}}}, ::DataStreams.Data.Schema{true,Tuple{Int64,Int64,Int64}}, ::Int64, ::Tuple{Base.#identity,Base.#identity,Base.#identity}, ::DataStreams.Data.##15#16, ::Array{Any,1}, ::Type{Ref{(:x, :y, :z)}}) at /home/mhu027/.julia/v0.6/DataStreams/src/DataStreams.jl:614
 [5] #stream!#17(::Bool, ::Dict{Int64,Function}, ::Function, ::Array{Any,1}, ::Array{Any,1}, ::Function, ::CSV.Source{Base.AbstractIOBuffer{Array{UInt8,1}},Void}, ::Type{DataFrames.DataFrame}) at /home/mhu027/.julia/v0.6/DataStreams/src/DataStreams.jl:490
[...]

PR #173, opened 2018-03-03
Improve handling of missing values

### LINKED ISSUE #153: Assume columns can have missing values by default
The current behavior of _CSV.jl_ is to scan columns until ```row_for_type_detect``` iterations have passed and infer their type.

However, for large files with few null cells, this force to set ```row_for_type_detect``` to the maximum number of rows, as the heuristic cannot detect if the last row of the file might contain null cells for instance.

--------

### Solution

One workaround would be to assume that rows can contain null cells by default, except if the user disable it. This was the behavior in ```DataFrames.readtable``` for instance.

Another would be to have a manual flag in ```CSV.Options``` to assume that all rows can contain null values.

nalimilan:
Replace the `nullable::Union{Bool,Missing}` argument with `allowmissing::Symbol`, with possible choices `:all`, `:auto` and `:none`, and make `:all` the default to avoid errors when the first missing value appears beyond rows used for type detection. Also avoid overriding types specified explicitly via the types argument,
which allows for more flexibility.

While deprecating `nullable`, also deprecate `null` in favor of `missingstring` for consistency with `truestring` and `falsestring`.

Fixes #153, https://github.com/JuliaData/CSV.jl/issues/159.


quinnj:
I went ahead and resolved the conflicts here; I'm seeing some weird failures on 0.7 where the `missingstring` sometimes disappears (i.e. when we try to access it during reading, it's empty even though we set it upon construction). Still need to do some more debugging, but the rest looks good.

nalimilan:
I've added a commit to get tests to pass on 0.6. They also pass locally on 0.7 with DataStreams master. I haven't seen the bug you mention, though. Good to merge?

nalimilan:
Bump! OK to merge?

andreasnoack:
See https://github.com/JuliaIO/JSON.jl/commit/f05dd3d26976ae8310cfb512cf0106505e95b51b#commitcomment-28964065. It also applies here.

PR #174, opened 2018-03-04
Use Int64 by default even on 32-bit

nalimilan:
Int32 overflows too easily, which is annoying.


nalimilan:
Yay, this fixed CI on 32-bit!

Issue #168, opened 2018-02-12
-Inf can not be parsed as Float64


Wikunia:
My csv has the following line:
`nuclear14,1563,1227,576,0,602,Min,-Inf`

If I have this header formats: 
`types=[String,Int64,Int64,Int64,Int64,Int64,String,Float64]`

it doesn't work and gives this error:

```
ERROR: LoadError: CSV.ParsingException("error parsing a `Float64` value on column 8, row 283; encountered '-'")
```

there doesn't seem to be an error with `Inf` instead of `-Inf`

nalimilan:
Looks like we need to change something here to take into account the possibility that a `'-'` character is followed by `"Inf"`: https://github.com/JuliaData/CSV.jl/blob/127041e862fad33d2bb4f0a30c930271ad87e512/src/float.jl#L94

Do you want to give it a try?

Wikunia:
Sry, how can I give it a try?

nalimilan:
I don't know, I just meant investigate how the code could be adapted to support this case.

Wikunia:
Ah okay. Sure I can give it a try later today.

jondea:
Also breaks on -nan

Wikunia:
@nalimilan sorry I'm not really sure how to implement it by myself.
@jondea Is there any use-case for minus not a number?

jondea:
@Wikunia, the bit representations are different and some languages/environments print them differently, see [Wikipedia page on not a number](https://en.m.wikipedia.org/wiki/NaN#Display). 

> Since, in practice, encoded NaNs have a sign, a quiet/signaling bit and optional 'diagnostic information' (sometimes called a payload), these will often be found in string representations of NaNs, too, for example: -NaN

As far as the logic goes, after discovering a minis sign, it needs to also check whether the next character is a number or a letter (or something along those lines). Currently it assumes minus sign implies number.

nalimilan:
I have a fix at https://github.com/JuliaData/CSV.jl/pull/176.

PR #176, opened 2018-03-04
Allow leading minus or plus sign before NaN and Inf

nalimilan:
Fixes https://github.com/JuliaData/CSV.jl/issues/168.

### DIFF
--- a/src/float.jl
+++ b/src/float.jl
@@ -90,15 +90,16 @@ end
 function parsefield(io::IO, ::Type{T}, opt::CSV.Options, row, col, state, ifnull::Function) where {T <: Union{Float16, Float32, Float64}}
     mark(io)
     @checknullstart()
-    negative = false
+    minussign = plussign = false
     if b == MINUS # check for leading '-' or '+'
-        c = peekbyte(io) 
+        minussign = true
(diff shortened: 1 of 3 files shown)


src/float.jl
```diff
@@ -133,7 +138,7 @@ function parsefield(io::IO, ::Type{T}, opt::CSV.Options, row, col, state, ifnull
             (!(b == LITTLEN || b == BIGN) || eof(io)) && (reset(io); b = readbyte(io); @goto checknullend)
             b = readbyte(io)
             !(b == LITTLEF || b == BIGF) && (reset(io); b = readbyte(io); @goto checknullend)
-            result = T(Inf)
+            result = Inf
```
quinnj: I think we still want these to be `T(Inf)`, right?

nalimilan: Woops, right, these are remnants of debugging tests. I've reverted them.

quinnj:
Thanks for this @nalimilan! I'm digging in to the 0.7 test failures locally.

nalimilan:
Thanks. FWIW, I've just tagged a CategoricalArrays release which passes tests on 0.7, if that can help.

PR #185, opened 2018-04-02
Use default_format from Dates

omus:
Avoids duplicating code and improves support for `ZonedDateTime` instances from TimeZones.jl.


omus:
Only Julia 0.7 tests are failing which is the same as the master branch.

PR #186, opened 2018-04-02
Set WeakRefStrings minimum version to v0.4.1

omus:
Alternatively we could just bump up the minimum version of WeakRefStrings that is required.


quinnj:
Yeah, I think I'm inclined to just bump the WeakRefStrings version.

omus:
Modified PR to change the minimum required version of WeakRefStrings to v0.4.1

quinnj:
Thanks.

PR #187, opened 2018-04-02
Correct use of non-field missingstringcheck

omus:
Mistake introduced in 4ca71c5.


quinnj:
Thanks!

nalimilan:
Woops. So that means it's not covered by tests? :-/

omus:
> Woops. So that means it's not covered by tests? :-/

Correct, there are definitely some gaps in test coverage.

PR #190, opened 2018-04-04
Support transforms as AbstractDict

omus:
Somewhat requires DataStreams.jl changes https://github.com/JuliaData/DataStreams.jl/pull/64. Without the DataStreams.jl changes passing anything but a `Dict` will fail.

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -334,13 +334,13 @@ sq1 = CSV.read(source, SQLite.Sink, db, "sqlite_table")
 """
 function read end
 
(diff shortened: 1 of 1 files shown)


src/CSV.jl
```diff
@@ -1,6 +1,7 @@
 __precompile__(true)
 module CSV
 
+using Base: Callable
```
nalimilan: Is this related? Otherwise, looks good.

omus: I believe I anticipated I would need this and then forgot to remove it. I'll take it out.

omus:
Note that `AbstractDict` was added to Compat.jl in v0.41 which happens to be the version of Compat already required.

PR #194, opened 2018-04-10
fix container type bug for `weakrefstrings=false`

### LINKED ISSUE #193: MethodError when attempting to populate `columntypes` in `Source`
```julia
v = Vector{DataType}(undef, 1)
v[1] = Union{Missing,Int}
```
```
ERROR: MethodError: Cannot `convert` an object of type Type{Union{Missing, Int64}} to an object of type DataType
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:123
Stacktrace:
 [1] setindex!(::Array{DataType,1}, ::Type, ::Int64) at ./array.jl:699
 [2] top-level scope
```
This is causing an error to get thrown [here](https://github.com/JuliaData/CSV.jl/blob/master/src/Source.jl#L207).  I'm not quite sure how this wasn't caught by testing.

PR incoming.

ExpandingMan:
Fix #193.

### DIFF
--- a/src/Source.jl
+++ b/src/Source.jl
@@ -198,7 +198,7 @@ function Source(;fullpath::Union{AbstractString,IO}="",
         autocols = Int[]
     end
     if !weakrefstrings
-        columntypes = [(T !== Missing && Missings.T(T) <: WeakRefString) ? substitute(T, String) : T for T in columntypes]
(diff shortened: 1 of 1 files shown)


ExpandingMan:
Failures on 0.7 due to old DataStreams version.

PR #203, opened 2018-05-04
RFC: Restart parsing with a wider type when CSV.read fails

nalimilan:
Introduce a new `ValueException` type to distinguish errors due to an incorrect value from errors due to an incorrect file structure. Use its fields to widen type of the offending column in case of error and restart parsing.

This is a very preliminary implementation to get feedback on the approach. I think it's a good solution to avoid throwing errors which are terribly confusing/annoying for users (no other software does that AFAIK). Of course it can only work 1) for `CSV.read` and not `CSV.source`, since DataStreams needs to know the types beforehand, and 2) when the sink is a type rather than an object. But this covers the most typical use case, where CSV.jl is currently less practical to use than other software.

Areas to yet improve:
- [ ] Widen `Int` columns to `Float64` before using `String`.
- [ ] Take into account a user-provided `types` argument.
- [ ] Figure out whether it's OK to always call `seekstart`, or whether some streams don't support it.
- [ ] Improve logging: should we print something at all when restarting? more details about the problematic column?
- [ ] Add a `restart=true` keyword argument to allow disabling the new behavior.
- [ ] Also support widening columns which initially do not allow for missing values when `allowmissing=:auto`.

For a simple test of the new behavior: `CSV.read(IOBuffer("X\n1\na"), rows_for_type_detect=1)`.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -9,6 +9,25 @@ struct ParsingException <: Exception
     msg::String
 end
 
+struct ValueException <: Exception
+    T::Type
+    v::Any
+    row::Int
+    col::Int
+end
+
+function Base.showerror(io::IO, e::ValueException)
+    if e.T === Missing
+        print(io, "encountered non-missing value for a missing-only column on row = $row, col = $col: '$v'")
+    elseif e.T <: AbstractFloat && e.v isa Integer
+        print(io, "error parsing a `$(e.T)` value on column $(e.col), row $(e.row): ",
+                  "exponent out of range: $(e.exp)")
+    else
+        print(io, "error parsing a `$(e.T)` value on column $(e.col), row $(e.row): ",
+                  "encountered '$(e.v)'")
+    end
+end
+
 const RETURN  = UInt8('\r')
 const NEWLINE = UInt8('\n')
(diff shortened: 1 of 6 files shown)


quinnj:
I actually want to support this in DataStreams itself and have some preliminary work in that direction.

nalimilan:
Cool. What kind of solution to you have in mind?

nalimilan:
After thinking more about this, the best solution seems to involve a combination of this PR and of your DataStream changes to allow widening types on the fly:
- When we encounter a non-numeric character in a column which is supposed to be numeric, we cannot just widen the type and continue: numeric values which have already been streamed cannot just be converted to string, they need to be read again (e.g. `001` differs from `1` as a string). So the only solution in that case is this PR.
- When we encounter a missing value in a column which isn't supposed to allow them, we can widen the type and continue. In that case your DataStream changes would be useful since we could stop using `allowmissing=:all` by default, and instead switch to `:auto`. That would be more practical for people who don't have missing values in their data (and who often find missing values confusing).

quinnj:
We now do perfect typing w/o needing to restart w/ #395

Issue #180, opened 2018-03-15
Julia crashes with "Bus error: 10" after CSV.write call


crbinz:
`test.csv` contains: 
```
A,B,C,D,E
12345,5,2018-03-10T19:48:06.0,abcdefg,hijklmn
```
Then:
```julia
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.2 (2017-12-13 18:08 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin14.5.0

julia> using CSV, DataFrames

julia> db = CSV.read("test.csv", dateformat=DateFormat("yyyy-mm-dd HH:MM:SS"), nullable = true);

julia> open("test.csv","w") do fh
       CSV.write(fh,db)
       end
Bus error: 10
```

```julia
julia> Pkg.status("CSV")
 - CSV                           0.2.2
```

Edit: the crash does not occur if I set `weakrefstrings=false`.

nalimilan:
I can reproduce on Linux. It also happens without `nullable=true`.

davidanthoff:
Isn't this kind of expected if you use ``weakrefstrings=true`` and memory mapped files? The arrays in ``db`` will reference memory locations that are directly mapped onto ``test.csv`` on disc. If you now start to change the file on disc, all those pointers to the various weak ref strings in ``db`` will point to who-knows-what.

nalimilan:
I'd have imagined `open("test.csv","w")` would have created a new copy of the file and that the OS would have kept the old file around until we close the handle. But apparently the file is modified in place.

Anyway this direct modification of the file is indeed possible and inevitably leads to corruption. We should probably avoid using both `weakrefstrings=true` and `use_mmap=true` by default: it's fine to use `mmap` to speed up parsing, but once it's done we shouldn't rely on the file being intact for the whole life of the data frame (and on Windows the file is locked, which isn't convenient either). I wonder what's the best default combination though: `weakrefstrings=false, use_mmap=true` or `weakrefstrings=true, use_mmap=false`? The former would have the advantage that we would return standard `Vector{String}` arrays, which are easier to understand for users.

[...]

PR #204, opened 2018-05-10
Intern strings by default instead of using WeakRefString

nalimilan:
`WeakRefString`s are dangerous in interaction with `use_mmap=true` since crashes can happen
if the file is modified during the lifetime of the resulting `DataFrame` (https://github.com/JuliaData/CSV.jl/issues/180). String interning can also be more efficient when the proportion of unique strings is small. Finally, returning plain `Vector{String}` columns is more user-friendly.

Ideally, DataFrames could take advantage of the fact that strings are interned to speed up grouping, which would allow using `categorical=false` by default.

A basic benchmark on 0.6 appears to indicate that interning is the fastest approach (not sure why the number of allocations is higher with `WeakRefString`):
```julia
julia> @btime CSV.read("test/test_files/Fielding.csv", strings=:intern, categorical=false, types=Dict("GS"=>Union{Int, Missing},"PO"=>Union{Int, Missing},"A"=>Union{Int, Missing},"E"=>Union{Int, Missing},"DP"=>Union{Int, Missing},"PB"=>Union{Int, Missing},"InnOuts"=>Union{Int, Missing},"WP"=>Union{Int, Missing},"SB"=>Union{Int, Missing},"CS"=>Union{Int, Missing},"ZR"=>Union{Int, Missing}), rows_for_type_detect=1);
  615.896 ms (10589390 allocations: 195.60 MiB)

julia> @btime CSV.read("CSV/test/test_files/Fielding.csv", strings=:weakref, categorical=false, types=Dict("GS"=>Union{Int, Missing},"PO"=>Union{Int, Missing},"A"=>Union{Int, Missing},"E"=>Union{Int, Missing},"DP"=>Union{Int, Missing},"PB"=>Union{Int, Missing},"InnOuts"=>Union{Int, Missing},"WP"=>Union{Int, Missing},"SB"=>Union{Int, Missing},"CS"=>Union{Int, Missing},"ZR"=>Union{Int, Missing}), rows_for_type_detect=1);
  675.738 ms (12563641 allocations: 255.85 MiB)

julia> @btime CSV.read("test/test_files/Fielding.csv", strings=:raw, categorical=false, types=Dict("GS"=>Union{Int, Missing},"PO"=>Union{Int, Missing},"A"=>Union{Int, Missing},"E"=>Union{Int, Missing},"DP"=>Union{Int, Missing},"PB"=>Union{Int, Missing},"InnOuts"=>Union{Int, Missing},"WP"=>Union{Int, Missing},"SB"=>Union{Int, Missing},"CS"=>Union{Int, Missing},"ZR"=>Union{Int, Missing}), rows_for_type_detect=1);
  725.674 ms (10589390 allocations: 195.60 MiB)
```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,5 +4,5 @@ DataFrames 0.11.0
 WeakRefStrings 0.4.1
 InternedStrings 0.6.0
 CategoricalArrays 0.3.0
-Compat 0.41.0
+Compat 0.50
 Missings
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,6 +2,7 @@ julia 0.6
 DataStreams 0.3.0
 DataFrames 0.11.0
 WeakRefStrings 0.4.1
+InternedStrings 0.6.0
 CategoricalArrays 0.3.0
 Compat 0.41.0
 Missings
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -1,7 +1,7 @@
 __precompile__(true)
 module CSV
 
-using DataStreams, WeakRefStrings, Missings, CategoricalArrays, DataFrames
+using DataStreams, WeakRefStrings, InternedStrings, Missings, CategoricalArrays, DataFrames
 
 using Compat, Compat.Mmap, Compat.Dates
 
@@ -55,6 +55,7 @@ Keyword Arguments:
  * `decimal::Union{Char,UInt8}`: character to recognize as the decimal point in a float number, e.g. `3.14` or `3,14`; default `'.'`
  * `truestring`: string to represent `true::Bool` values in a csv file; default `"true"`. Note that `truestring` and `falsestring` cannot start with the same character.
  * `falsestring`: string to represent `false::Bool` values in a csv file; default `"false"`
+ * `internstrings`: whether strings should be interned, rather than creating a new object for each string field; default `true`
 """
 struct Options{D}
     delim::UInt8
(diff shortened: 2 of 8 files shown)


nalimilan:
Yes, it's really cool how simple this approach is. I agree `WeakRefStringArray`/`StringArray` are useful too, but they are better kept as options given that they only work well when you can ensure the backing file is preserved.

Fixing the 0.7 failures doesn't seem to be hard, but while doing that I bumped on a strange fact:
```julia
julia> x = UInt8['a'];
julia> y = UInt8['a'];

julia> String(x) === String(y)
true

julia> x === y
false
```

So it looks like Julia 0.7 interns strings by default? That would mean most of that PR can be dropped once 0.7 is out.

quinnj:
Yeah, not "interned by default", but it's more the result of all the alloc-elim pass work; i.e. the compiler is smarter at detecting when things are the same and avoiding unnecessary allocations when objects are "the same" (`===`). Interning strings will still be useful for the foreseeable future.

nalimilan:
> Yeah, not "interned by default", but it's more the result of all the alloc-elim pass work; i.e. the compiler is smarter at detecting when things are the same and avoiding unnecessary allocations when objects are "the same" (===). Interning strings will still be useful for the foreseeable future.

IIUC what Stefan said on Slack (at #strings), the new behavior of `===` is just a semantic change, it doesn't actually mean that the compiler avoids allocations when possible. And indeed `pointer` still shows that objects are actually different.

I've added a commit which fixes tests on 0.7 with https://github.com/JuliaString/InternedStrings.jl/pull/15 (except for a failure which is already present on master).

nalimilan:
Tests pass locally on 0.7 with DataStreams master. Merging.

PR #206, opened 2018-05-14
Update Compat requirement

andreasnoack:
See https://github.com/JuliaData/CSV.jl/pull/173#issuecomment-388800896


nalimilan:
What function requires 0.59? AFAICT my PR didn't use `occursin`, right?

We really need a way to run the tests with the minimal required versions, rather than with the highest ones.

andreasnoack:
> What function requires 0.59?

It's `Array(undef, ...)` in https://github.com/JuliaData/CSV.jl/pull/173/files

Issue #86, opened 2017-06-28
currently not parsable in CSV.read() but parses fine in DataFrame's readable()


tcovert:
I have an example file which reads fine in `readtable()` from `DataFrames` but will not read using `CSV.read()': 
https://www.dropbox.com/s/hrutgyidhojujv9/WellIndex_20160811.csv?dl=0

Since I am able to get this data into Julia using DataFrames I am personally in good shape, but I understand that eventually the plan is to deprecate `readtable()` in favor of this package, so it would be cool if it is eventually possible to parse files like this.

Thanks in advance if you are able get `CSV.read()` to read this.

posted [here](https://github.com/JuliaComputing/TextParse.jl/issues/19) last month as well...

nalimilan:
Could you reduce the file to a smaller test case? The Dropbox link made my browser hang for seconds before I could download the file. It would be helpful to remove any rows and columns which are not needed to reproduce the problem.

tcovert:
The error message I get is kind of cryptic, so its hard to know which rows/cols are triggering the error:
```
julia> wi = CSV.read("WellIndex_20160811.csv", DataTable)
ERROR: CSV.CSVError("Unexpected start of quote (34), use \"9234\" to type \"34\"")
Stacktrace:
 [1] readsplitline!(::Array{CSV.RawField,1}, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::UInt8, ::UInt8, ::UInt8, ::Base.AbstractIOBuffer{Array{UInt8,1}}) at /Users/tcovert/.julia/v0.6/CSV/src/io.jl:89
 [2] #Source#12(::String, ::CSV.Options, ::Int64, ::Int64, ::Array{DataType,1}, ::Bool, ::Bool, ::Int64, ::Int64, ::Int64, ::Bool, ::Type{T} where T) at /Users/tcovert/.julia/v0.6/CSV/src/Source.jl:137
 [3] (::Core.#kw#Type)(::Array{Any,1}, ::Type{CSV.Source}) at ./<missing>:0
[...]

Issue #166, opened 2018-02-08
Change default escapechar='"'


heaven00:
Datafile = train.csv from [here](https://www.kaggle.com/c/jigsaw-toxic-comment-classification-challenge/data)

**Versions**
Julia 0.6.2
CSV 0.2.1+

**Description**
While reading the file its not able parse `\"` because of the default escapechar='\\' in the textual content within the 2nd column .i.e. `comment_text`
**Suggested solution**
set `escapechar='"'` which worked on julia 0.7 and julia 0.6 but only on @quinnj machine.

I am running Manjaro which is an archlinux flavour, any help would be appreciated :)

**Workaround**
So i tried using the `readdlm` that worked nicely for me 
```DataFrame(readdlm("data/train.csv", ','; skipstart=1), [:id, :comment_text, :toxic, :severe_toxic, :obscene, :threat, :insult, :identity_hate])```

nalimilan:
TextParse.jl/CSVFiles.jl, R, dplyr/readr and Pandas use `""` to represent a quote inside a quoted field by default, which is consistent with the "[CSV standard](https://tools.ietf.org/html/rfc4180)".

@quinnj Do you confirm you're OK with switching to `escapechar='"'` by default? Though other implementations seem to handle double quotes via a separate argument, but it's not clear to me whether there's an advantage to doing so.

nalimilan:
An interesting reference is W3C's [Model for Tablular Data on the Web](https://www.w3.org/TR/tabular-data-model/#parsing). It recommends using the quote as the escape char by default, so that double quotes are interpreted as single quotes inside quoted fields. Their escape char only escapes quotes, not other sequences. So maybe by default we should only treat double quotes as single quotes, but not interpret other escape sequences in any special way.

nalimilan:
See https://discourse.julialang.org/t/problem-related-to-csv-read-and-strings/9595/8 for more examples.

PR #215, opened 2018-06-08
Is escapechar broken?

gustafsson:
The documentation states:

```
    •    quotechar::Union{Char,UInt8}: the character that indicates a quoted field that may contain the delim or newlines; default '"'
      
    •    escapechar::Union{Char,UInt8}: the character that escapes a quotechar in a quoted field; default '\'
```

I interpret this as the following csv file
```
x
"a\"b"
```
should be parsed as `DataFrame(x="a\"b")`. But this isn't the case. Am I missing something or is something broken?

Similarly the test of #86 didn't test that `escapechar='"'` transformed `""` to `"`. The test looks broken to me.


nalimilan:
See also https://github.com/JuliaData/CSV.jl/issues/166.

gustafsson:
Isn't #166 a discussion about what the default should be?

My apologies for being so blunt, but this PR is more about that I wasn't able to find at least one example file where at least one `escapechar` works (including the default).

---

Let's take another example from: https://discourse.julialang.org/t/problem-related-to-csv-read-and-strings/9595/8

```
julia> open("/tmp/tst.csv", "w") do f write(f, IOBuffer("a,b,c\r\n104652,\"Thanks  \\\",a\r\n")) end

shell> cat /tmp/tst.csv
a,b,c
104652,"Thanks  \",a

julia> CSV.read("/tmp/tst.csv", quotechar='"')
0×3 DataFrames.DataFrame
```

This seems to be another issue with empty last lines. But

```
julia> open("/tmp/tst.csv", "w") do f write(f, IOBuffer("a,b,c\r\n104652,\"Thanks  \\\",a")) end

shell> cat /tmp/tst.csv
a,b,c
104652,"Thanks  \",a
julia> CSV.read("/tmp/tst.csv", quotechar='"')
1×3 DataFrames.DataFrame
│ Row │ a      │ b             │ c       │
├─────┼────────┼───────────────┼─────────┤
│ 1   │ 104652 │ Thanks  \\",a │ missing │
```

nalimilan:
> Isn't #166 a discussion about what the default should be?

Yeah, I just mentioned it because we wondered how it should work when that issue was filed. But I have no idea how broken it is.

quinnj:
Alright, I've finally got a handle on this and I'm a bit embarrassed. Unfortunately, I never took the time to fully think through how escape characters work and only ever considered their implications in escaping delimiters/newlines, but not in how they should be removed in the final data. I _should_ have got this because we do this [exact same thing](https://github.com/quinnj/JSON2.jl/blob/0a37062e539b81a28a4e9e0e7661e43315ed424e/src/strings.jl#L124) in JSON2.jl. I just tagged a release of Parsers.jl that ensures we `unescape` strings if escape characters are found while parsing. Sorry for the hassle and confusion everyone.

PR #218, opened 2018-06-19
Require Julia 0.7-alpha to run Femtocleaner

nalimilan:


nalimilan:
@Keno Looks like Femtocleaner hangs again? :-)

PR #240, opened 2018-08-14
fixed 0.7 deprecation warning

MLackner:
specifically the `replace` function


nalimilan:
Thanks. Could you revert the addition of Manifest.toml (which shouldn't generally be committed)?

MLackner:
Okay... Removed it :)

nalimilan:
Thanks. Though I guess we should remove REQUIRE if we add Project.toml?

MLackner:
Frankly, I have no idea at which point the REQUIRE file should be replaced by the Project.toml file. I just put the Project.toml file with a genereic uuid in there because my tests were complaining about missing dependencies in CSV.jl. So this was just to get my code working again with 0.7.

quinnj:
Sorry, I pushed a fix separate from this. Thanks for the contribution though!

quinnj:
Also, just as FYI: packages still require REQUIRE files until METADATA.jl has been fully transitioned to expect Project.toml files. It's also not discouraged to commit Manifest.toml files, but package maintainers need to be careful when committing to ensure the Manifest represents a valid environment for any user/maintainer (i.e. no local dev paths, etc.).

PR #243, opened 2018-08-21
Fix escaping of strings that include escape characters

quinnj:


quinnj:
This is in master already.

PR #246, opened 2018-08-25
Refactor csv reading

quinnj:
- [x] Add some more tests around new functionality
- [x] Add some precompile statements to speed up first runs
- [x] Figure out what to do about `transpose=true`
- [ ] Update docs


src/CSV.jl
```diff
@@ -1,139 +1,266 @@
...
+    escapechar::Union{UInt8, Char}='\\',
+    dateformat::Union{String, Dates.DateFormat, Nothing}=nothing,
+    decimal::Union{UInt8, Char, Nothing}=nothing,
+    truestrings::Union{Vector{String}, Nothing}=nothing,
+    falsestrings::Union{Vector{String}, Nothing}=nothing,
+    # type options
+    types=nothing,
+    typemap::Dict=Dict{Type, Type}(),
+    allowmissing::Symbol=:all,
+    categorical::Bool=false,
```
nalimilan: If you break the API, it would be a good time to replace `categorical` with `strings=:categorical`, as these two arguments are not really independent. But I see the `strings` argument isn't there anymore. Does that mean one cannot avoid interning strings?

BTW, should add one indentation space to get correct alignment.

quinnj: Yes, strings are interned by default in the Parsers.jl package. I don't see any reason not to always intern strings? And for WeakRefStrings, users can state them specifically in `types` or there's a new `typemap` keyword argument to swap a type for type for the whole file.

nalimilan: It's maybe not essential, but it happens that a CSV file contains strings which never appear twice, e.g. names, IDs or free-form answers to a questionnaire. But I guess this would need to be decided on a per-column basis, so maybe we don't need the `strings` argument after all. Another use case would be to choose a custom string, for example to use a special encoding, but that can be added later.

At any rate I'd remove the `length(levels[col]) / sum(values(levels[col])) < .67` hack: better always return categorical or non-categorical when the user passes either `true` or `false`, and only use the trick when the user passes a real (or even remove it).

PR #259, opened 2018-08-31
Add a couple extra tests and cleanup code a bit

quinnj:


src/typedetection.jl
```diff
@@ -40,34 +40,36 @@ initialtypes(T, t::Dict{Symbol, V}, names) where {V} = Any[get(t, nm, T) for nm
 initialtypes(T, t::Dict{Int, V}, names) where {V} = Any[get(t, i, T) for i = 1:length(names)]
 initialtypes(T, t::Vector, names) = length(t) == length(names) ? collect(Any, t) : throw(ArgumentError("length of user provided types ($(length(t))) does not match length of header (?$(length(names)))"))
 
-Z(cv, p, moe) = (cv^2 * p * (1 - p)) / moe^2
-# 99% CI, 50% proportion, 2.5% margin of error
-const X = Z(2.58, 0.5, 0.025)
-samplesize(N) = ceil(Int, (N * X) / (X + N - 1))
+struct Fib{N}
```
quinnj: @nalimilan, I've replaced the "sample size" type detection code w/ this new fibonacci sequencer. This follows the strategy of Base Array hashing in that it will take a sequence of N rows (currently N = 1024, but I wonder if we could make that configurable somehow), skipping `fib` # of elements in between each iteration. So it will take the first 1:1024 rows, then start skipping every other row, then every 3rd row, etc. This seems to work well on all the files we currently cover and we end up using more rows for larger datasets than the sample size strategy, which is good because the overhead is less in those cases. Overall, I'm happy with this approach.

nalimilan: That's nice, but thinking about it again, I wonder what's the justification for skipping more and more rows. This assumes the first rows are more interesting, which isn't totally obvious. In particular, the last rows are likely to differ from others, e.g. if they are sorted and missing values are at the end. It would probably be better to take a uniform proportion of rows over the whole file, wouldn't it?

So maybe it would make more sense to use some rule to choose the proportion of rows to sample (maybe using something like `N = log(nrow)`), and then parse e.g. 50 rows every `nrow/N/50` rows?

PR #254, opened 2018-08-30
Fix #71 by adding a keyword argument to ignore repeated delimiters

quinnj:

PR #266, opened 2018-09-03
Rename ignore_repeated_delimiters argument to ignorerepeated

nalimilan:
That's consistent with other arguments, which don't use underscore, and it's easier to type.

See https://github.com/JuliaData/CSV.jl/pull/254#discussion_r214061557.


sylvaticus:
"mergedelimiters" as in Libre/OpenOffice ?

sylvaticus:
I would still prefer "mergedelim", as "ignorerepeated" doesn't say.. what the adjective "repeated" refers to.
I could make a pull request, but before I would like to know if there is consent for that or.. it is just me ;-)

nalimilan:
I'd be fine with `mergedelim` too.

PR #268, opened 2018-09-05
Renamed `ignorerepeated` to `mergedelim`

sylvaticus:


quinnj:
Hmmm, I don't think I'm a fan of `mergedelim` over `ignorerepeated`. I also don't _really_ care, so I'll let others chime in.

nalimilan:
I don't have a strong opinion either. `mergedelim` has the advantage that it mentions "delim", which probably makes it easier to find. But "merge" isn't totally obvious.

Do other software have a name for this (beyond LibreOffice)?

quinnj:
Thanks for the PR @sylvaticus, but it doesn't look like there's much support for renaming right now. We can discuss this more if more people think it should be changed.

PR #271, opened 2018-09-08
types in CSV.file docstring

jannefiluren:
I changed, hopefully improved, the docstring a bit, and also added some more info about allowed input types.

Short questions:

- limit: is that the maximum number of data rows to read of the file (limit minus eventual header rows)?

-  types: how does one use a Dict of types to be used for column types?

- How does one use UInt8 as input?

Short comment:

I moved `use_mmap` further down since that does not seem like to most important argument. What does this actually do?

I will look at the test an try to augment the index.md file.


jmgnve:
I'll update the pr soon.

- I think I kept `use_mmap` in the "File layout options" but put it at the end. Is that ok?

- I will remove all `UInt8` stuff from the docstring, and perhaps add a note inte `docs` I am working a bit on.

jmgnve:
I hope this is okay. I used your comments, and `use_mmap` is still in the list but a bit further down. It somehow didn't work to amend to my previous commit, so that is why there is a weird merge. I have not yet a black belt in git...

src/CSV.jl
```diff
@@ -175,9 +175,9 @@ Base.propertynames(row::Row{F}) where {F <: File{NamedTuple{names, T}}} where {n
 """
     CSV.File(source::Union{String, IO}; kwargs...) => CSV.File
 
-Read a csv input (a filename given as a String, or any other IO source), returning a `CSV.File` object.
```
nalimilan: The convention is to use the imperative form. See https://docs.julialang.org/en/latest/manual/documentation/

PR #286, opened 2018-09-20
Refactor CSV.write internals a bit to ensure we always escape things …

quinnj:
…properly

cc: @bkamins


bkamins:
Thanks! The only question is that if we use global `FILE_BUFFER` and `VALUE_BUFFER` that this will be not thread safe (which is OK for the time being but it should be noted in the documentation then). Or I am missing something?

bkamins:
I have just noticed that I have put a comment on green threads in a wrong thread as I was working on two things in parallel 😄. But I guess it did not confuse you.

The simplest option could be even to have a global lock per thread set to on when starting `CSV.write` and released when it finishes. In this way we have to lock only once per `CSV.write` call which should not be that bad. And I guess anyway the current design discourages using `CSV.write` in several green threads in parallel. Anyway - I am OK if you decide to whatever you think is best.

Issue #282, opened 2018-09-18
Validation feature request


bkamins:
I would like to put under consideration to add `validate` keyword to `CSV.File` that would turn on/off detailed validation of the file via `CSV.Validate`.

Additionally personally I would like to `validate` default to `true` as people should opt-out of validation if they need performance.

bkamins:
Just to put an example of a problem I want to solve is that currently `CSV.read` has the following behavior:
```
julia> CSV.read(IOBuffer("a,b,c\n1\n2\n3"))
3×3 DataFrames.DataFrame
│ Row │ a       │ b       │ c       │
│     │ Int64⍰  │ Int64⍰  │ Int64⍰  │
├─────┼─────────┼─────────┼─────────┤
│ 1   │ 1       │ 2       │ 3       │
│ 2   │ missing │ missing │ missing │
│ 3   │ missing │ missing │ missing │
```
which is caught only by `CSV.validate`.

bkamins:
CC @nalimilan

quinnj:
I can take a look at that specific issue, but I'd have preferred opening a new issue, since it's really a separate issue than the discussion around `validate`.

quinnj:
@bkamins can you check your version please?
```julia
julia> using CSV, Test, Tables
[ Info: Recompiling stale cache file /Users/jacobquinn/.julia/compiled/v1.0/CSV/HHBkp.ji for CSV [336ed68f-0bac-5ca0-87d4-7b16caf5d00b]

julia> CSV.read(IOBuffer("a,b,c\n1\n2\n3"))
3×3 DataFrames.DataFrame
│ Row │ a      │ b       │ c       │
│     │ Int64⍰ │ Missing │ Missing │
├─────┼────────┼─────────┼─────────┤
│ 1   │ 1      │ missing │ missing │
│ 2   │ 2      │ missing │ missing │
│ 3   │ 3      │ missing │ missing │

```

bkamins:
Confirmed - on `HHBkp` it works like you reported above. Seems that CSV.jl is a fast moving target 😄 - fortunately in a good direction.

nalimilan:
Cool. So what's the cost/benefit tradeoff of `validate=true`?

quinnj:
[...]

PR #292, opened 2018-09-22
Transition CSV.read to use new CSV.File machinery underneath

quinnj:
cc: @nalimilan @bkamins


bkamins:
Very nice. It strikes a good balance wrt. `DataFrame`. 👍 for removing transform in the long run.

The only issue I have, as noted in #282 is to:

1. add a keyword parameter allowing to enable validation of the file
2. Then actually we might differentiate the default value of this parameter between `CSV.File` (set to `false` - as this is something that users would use for performance) and `CSV.read` (set to `true` - as this is the option that newcomers would use - and it is better that they do not say that CSV.jl does not work properly)

(but this is a thing that independent and can be discussed later in #282 - I stress it because I feel it is important for new users switching to Julia ecosystem from R/Python)

quinnj:
Tests for transform shim added in 6126d7cc24cef1a1fd0d712766a032116c57160b

Issue #278, opened 2018-09-18
Keep CategoricalArray levels sorted


nalimilan:
This is a reminder that we should ensure the levels of `CategoricalArray` columns are always sorted during parsing. We currently keep the order in which levels were added, which is meaningless since it depends on which rows where used for type detection. But reordering levels only at the end of parsing would be a bit weird too since the relative order of existing levels would be changed, yet in streaming operations users could have used that order.

So the best solution is probably to sort levels everytime we call `push_level!`. This shouldn't have a large performance impact since integer reference codes are not affected.

PR #293, opened 2018-09-22
Fix and improve CategoricalArray handling

nalimilan:
See commit messages. This should definitely be merged before tagging a release, since `CategoricalArray` handling is actually quite broken now. Also I think it's a good time to change the behavior of `categorical=true`.

Closes https://github.com/JuliaData/CSV.jl/issues/278.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -599,7 +599,13 @@ end
         return missing
     else
         @inbounds pool = source.pools[col]
-        return get🐱(pool, str::Tuple{Ptr{UInt8}, Int})
+        str::Tuple{Ptr{UInt8}, Int}
+        i = get(pool, str, nothing)
+        if i === nothing
+            i = get!(pool, unsafe_string(str[1], str[2]))
+            issorted(levels(pool)) || levels!(pool, sort(levels(pool)))
+        end
+        return CatStr(i, pool)
     end
 end
 
--- a/src/typedetection.jl
+++ b/src/typedetection.jl
@@ -99,8 +99,8 @@ function detect(types, io, positions, parsinglayers, kwargs, typemap, categorica
         pools = Vector{CategoricalPool{String, UInt32, CatStr}}(undef, cols)
         for col = 1:cols
             T = types[col]
-            if length(levels[col]) / sum(values(levels[col])) < .67 && T !== Missing && Base.nonmissingtype(T) <: String
-                types[col] = substitute(T, CategoricalArrays.catvaluetype(Base.nonmissingtype(T), UInt32))
+            if T === String || T === Union{String, Missing}
+                types[col] = substitute(T, CatStr)
                 pools[col] = CategoricalPool{String, UInt32}(collect(keys(levels[col])))
             end
         end
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -35,7 +35,7 @@ struct File{transpose, columnaccess, I, P, KW}
     lastparsedcol::Base.RefValue{Int}
     lastparsedcode::Base.RefValue{Parsers.ReturnCode}
     kwargs::KW
-    pools::Vector{CategoricalPool{String, UInt32, CategoricalString{UInt32}}}
+    pools::Vector{CategoricalPool{String, UInt32, CatStr}}
     strict::Bool
 end
 
@@ -172,7 +172,13 @@ function File(source::Union{String, IO};
     end
 
     if types isa Vector
-        pools = CategoricalPool{String, UInt32, CatStr}[]
+        pools = Vector{CategoricalPool{String, UInt32, CatStr}}(undef, length(types))
+        for col = 1:length(types)
+            T = types[col]
+            if T !== Missing && Base.nonmissingtype(T) <: CatStr
+                pools[col] = CategoricalPool{String, UInt32}()
+            end
+        end
     else
         types, pools = detect(initialtypes(initialtype(allowmissing), types, names), io, positions, parsinglayers, kwargs, typemap, categorical, transpose, ref, debug)
         if allowmissing === :none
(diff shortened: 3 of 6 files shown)


src/deprecated.jl
```diff
@@ -599,8 +599,14 @@ end
         return missing
     else
         @inbounds pool = source.pools[col]
-        return get🐱(pool, str::Tuple{Ptr{UInt8}, Int})
```
nalimilan: :cry:

src/deprecated.jl
```diff
@@ -599,8 +599,14 @@ end
...
-    end
+        p, len = str::Tuple{Ptr{UInt8}, Int}
+        level = unsafe_string(p, len)
```
nalimilan: @quinnj Would it be more efficient to use a `WeakRefString` or something like this here? We just need to be able to hash the object and call `==` on it.

quinnj: See [here](https://github.com/JuliaData/Parsers.jl/blob/master/src/strings.jl#L19) for why this works. Can we use the old code here that checked `ht_keyindex2`; it's much faster than doing two `get` calls.

src/deprecated.jl
```diff
@@ -599,8 +599,14 @@ end
...
+        i = get(pool, level, nothing)
+        if i === nothing
+            i = get!(pool, level)
+            issorted(levels(pool)) || levels!(pool, sort(levels(pool)))
+        end
+        return CatStr(i, pool)    end
```
quinnj: What's with the `end` on this line?

src/typedetection.jl
```diff
@@ -99,8 +99,8 @@ function detect(types, io, positions, parsinglayers, kwargs, typemap, categorica
         pools = Vector{CategoricalPool{String, UInt32, CatStr}}(undef, cols)
         for col = 1:cols
             T = types[col]
-            if length(levels[col]) / sum(values(levels[col])) < .67 && T !== Missing && Base.nonmissingtype(T) <: String
-                types[col] = substitute(T, CategoricalArrays.catvaluetype(Base.nonmissingtype(T), UInt32))
+            if T === String || T === Union{String, Missing}
```
quinnj: This makes me worried though for String `id` columns; sure I want a low cardinality string column to be made Categorical, but for a highly unique column, does it really make sense?

src/deprecated.jl
```diff
@@ -599,8 +599,14 @@ end
...
     else
         @inbounds pool = source.pools[col]
-        return get🐱(pool, str::Tuple{Ptr{UInt8}, Int})
-    end
+        p, len = str::Tuple{Ptr{UInt8}, Int}
+        level = unsafe_string(p, len)
+        i = get(pool, level, nothing)
+        if i === nothing
+            i = get!(pool, level)
+            issorted(levels(pool)) || levels!(pool, sort(levels(pool)))
```
quinnj: Should we just sort them after insert? I can't imagine that levels will come in sorted very often, so it would avoid the extra `issorted` check. I wonder if we should also consider a sorting algorithm optimized for "nearly sorted"; probably not a big deal, but I know some algorithms have a toxic path when the input is almost sorted.

src/deprecated.jl
nalimilan: It's only repeated in a deprecated function.

src/deprecated.jl
nalimilan: Mmm, except that it didn't... :-/ The test I added failed, that's why I rewrote this.

Anyway, that's severe type piracy! I can try to make this work without type piracy using a custom wrapper struct if you want.

src/deprecated.jl
nalimilan: I'll fix it.

src/typedetection.jl
nalimilan: Not really, but people should think about what they do if they set `categorical=true`. We could always allow passing a number to choose a threshold, but I prefer providing simple options first. We've had reports by users who were confused by this clever logic.

It would also make sense to be able to combine `categorical=true` with e.g. `types=(:ID=>String)` to choose to keep `:ID` as a string column.

src/deprecated.jl
nalimilan: CategoricalArrays need to do some work in `levels!` which is likely to be much more costly than calling `issorted`, so I think it's more efficient that way. If we want to improve performance here, we could need to add an argument to the function which adds levels, and have `CategoricalPool` store a Boolean to remember whether levels are currently sorted.

Anyway, the number of levels is supposed to be much lower than the number of rows (that's the whole point of CategoricalArrays), so this shouldn't really matter.

src/deprecated.jl
quinnj: Personally, I think people are way too heretical about type piracy; this is far from severe. It's an obscure type that's rarely used; if someone is dealing w/ a `Tuple{Ptr{UInt8}, Int}`, they'd probably _want_ this behavior anyway.

I'd have to look into why it didn't work in this case.

nalimilan: OK, got it. Calling `push_level!` is not enough, that's an internal function which is supposed to be combined with an insertion into `invindex`. I don't like calling these non-public functions, so I've changed it to `get` and `get!`. We can always add an API to CategoricalArrays to avoid the double lookup, but for now it should be fast enough.

So I've reverted to passing the tuple directly for the lookup. But that type piracy should really be fixed. Also (and maybe more problematically), the `isequal` definition is vulnerable to hash collisions, which isn't acceptable when dealing with user data.

nalimilan: Why not use `WeakRefString` instead of a tuple?

quinnj: We can certainly use WeakRefString; that's the right call here.

src/typedetection.jl
nalimilan: See https://github.com/JuliaData/CSV.jl/pull/317.

PR #294, opened 2018-09-22
Enable CI on Julia 0.7

nalimilan:
Also allow failures on nightlies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,13 @@ os:
   - osx
 
 julia:
+  - 0.7
   - 1.0
   - nightly
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,9 +7,14 @@ os:
   - osx
 
 julia:
+  - 0.7
```
quinnj: Why 0.7? Everything works fine on 1.0, so I feel like this is just wasting CI.

nalimilan: To prove my point: tests fail on 0.7 under Windows... ;-)

Issue #295, opened 2018-09-22
On column name transformations


gdkrmr:
On current master you automatically transform column names so that the symbol representation in Julia is simpler (`:a_b` instead of `Symbol("a b")`).

I guess R did this for the same reason you changed the standard behaviour, i.e. to facilitate `df.column` notation, `df$column` in R. The `data.table`/`readr` packages changed the base `R` behaviour, the `data.table` way is much more predictable and therefore easier to use. 
Seeing these parallels and that the newer `R` packages reversed this, you may want to reconsider this change.

Other remarks:
- The current transform you use also fails if the column name consists solely of numbers (`"2017"` becomes `Symbol("2017")` on current master, base `R` used to solve this by prepending the letter `X`, e.g. `"2017"` becomes `X2017`).

- add the following functions:
```julia
function simplify_symbol(x)
    x |> String |> x -> replace(x, " " => "_") |> Symbol
end

function simplify_column_names!(x)
    names!(x, simplify_symbol.(names(x)))
end
```

quinnj:
Is there a request here? You mention that `data.table` does something different than base `R`, but you failed to explain what they do "more predictably". You also mention newer `R` packages reversed this, what does that mean? Which R packages? What did they reverse?

nalimilan:
I hadn't realized that neither data.table nor readr (which are the "newer packages" compared to base R) normalize column names. And AFAICT Pandas doesn't either. So we should probably do the same thing, with an argument to enable normalization.

gdkrmr:
> Is there a request here?

- This is not a real request, I think this is something that the package maintainers have to decide. My intention was to explain what happened in R so you can avoid similar mistakes in CSV.
- The most important thing is to be stable, changing everything from `:a_b` to `Symbol("a b")` is quite annoying, so you should take a decision and stick with it.

> You mention that data.table does something different than base R, but you failed to explain what they do "more predictably". You also mention newer R packages reversed this, what does that mean? Which R packages? What did they reverse?

sorry if this wasn't clear: base `R` normalizes column names so that the symbol representation is more readable (`df$a.b`), while all the newer packages (`data.table`, `readr`) do not (`df$'a b'`, these should be backticks).

gdkrmr:
I forgot to add:
- I don't really want to tell you what to do with this because I do not know the reasons behind `data.table` and `readr` not following standard R behavior.

[...]

PR #302, opened 2018-09-25
Don't normalize names by default

quinnj:


quinnj:
Based on discussions in https://github.com/JuliaData/CSV.jl/issues/295

PR #303, opened 2018-09-26
Adjust the columns threshold based on various sized benchmark results

quinnj:
Ok, ran some experiments with various sized files; obviously there are some shortcomings to the approach here: it's tied to the performance characteristics of my laptop, I only tested Float64 columns, etc, etc. But this is at least more informed than the arbitrary threshold that was previously chosen. We can always refine things further later on.
screen shot 2018-09-25 at 10 21 21 pm

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -371,3 +371,31 @@ function Base.getproperty(csvrow::Row{F}, ::Type{T}, col::Int, name::Symbol) whe
     end
     return r
 end
+
+
+# code for generating various-sized csv files
+# used to determine the row vs. column-access threshold
+function gencsv(rows, cols)
+    df = DataFrame([round.(rand(rows), digits=4) for _ ∈ 1:cols], Symbol.(["col$i" for i ∈ 1:cols]))
+    CSV.write("random_$(rows)_$(cols).csv", df)
+end
+
+function go(compile=true)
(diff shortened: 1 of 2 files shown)


nalimilan:
Out of curiosity, does the trade-off vary depending on the number of columns?

quinnj:
Yes, the x-axis in the chart there is number of _cells_, so rows * columns.

quinnj:
It's also what prompted [this change](https://github.com/JuliaData/Tables.jl/commit/87c74d28afd52d8f41a88e53381acaf7985735f5); where we compile type-specific for-loops instead of straight-line expressions in Tables.jl `eachcolumn`. Previously it only kicked in if the # of columns was less than 100, but in the chart, the compile times are very fast using the run-length encoding, which kicked in at the 1000 cells mark. Faster than even datasets with only 10 columns.

nalimilan:
What I wonder is whether for a given number of cells, the rows/columns ratio makes a difference. (That RLE trick is really interesting!)

quinnj:
I'm not sure the # of columns really makes a difference outside the dynamic compilation cost in the original chart; the original threshold, however, was only based on the # of rows, which didn't quite provide a great answer for really wide datasets where it ended up being faster to dynamically compile. Accounting for # of columns in the threshold helps.

PR #320, opened 2018-10-09
Fix docs to match actual default arguments

coreywoodfield:
Docs said the default argument for `escapechar` in `CSV.File` was `'\\'`, but it's actually `'"'`


quinnj:
Thanks!

Issue #324, opened 2018-10-13
using CSV became extremly slow


ufechner7:
With the latest version of CSV the load time on Julia 1.0.1 and Julia 0.7 increased 10 times:
```
  | | |_| | | | (_| |  |  Version 1.0.1 (2018-09-29)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using CSV
 26.338247 seconds (43.67 M allocations: 2.136 GiB, 5.79% gc time)
```
Same machine:
```
  | | |_| | | | (_| |  |  Version 0.6.2 (2017-12-13 18:08 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-pc-linux-gnu

julia> @time using CSV
  2.881148 seconds (2.54 M allocations: 146.385 MiB, 11.33% gc time)
```
It used to be faster on 0.7 then on 0.6.

Any idea?

nalimilan:
The codebase has changed a lot in the last few months, so it's probably just due to that.

ufechner7:
We are mainly do flight testing of wind drones. After each flight (they are usually short, like 10 min) we analyse the csv log file, which is not small, because we are logging about 120 values at 200 Hz. With Julia 0.6 loading a log file and preprocessing it takes about 20s on my laptop. I will create a more realistic test for loading CSV and loading the data to see how much slower the current version of CSV is compared to the 0.2.5 version that I used with Julia 0.6.

nalimilan:
Then it shouldn't make a difference for you whether compilation happens when loading the package or when reading a file for the first time. In the end the total time should be similar. You can easily check that by removing the call I linked to above.

quinnj:
The CSV.jl-induced regression is at least fixed on master now.

PR #325, opened 2018-10-15
don't force compiling on init

KristofferC:
There are many cases where CSV is loaded as a dependency for a package but is quite unlikely to be used at all in the session. The current strategy of compiling in `__init__` means that everyone has to pay that cost just to use CSV as a dependency. The fact that Julia has a compiler bundled is a strength of the language.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -242,9 +242,6 @@ include("validate.jl")
 include("transforms.jl")
 
 function __init__()
-    # read a dummy file to "precompile" a bunch of methods; until the `precompile` function
-    # can propertly store machine code, this is our best option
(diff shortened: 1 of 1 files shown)


nalimilan:
x-ref https://github.com/JuliaData/CSV.jl/issues/324.

quinnj:
But the reason we do this is because we've gotten reports for years about people complaining how slow the first `CSV.read` is. I think this is the current best solution for Julia not being able to fully precompile machine code.

quinnj:
It really depends on which one would potentially annoy @JeffBezanson or @vtjnash the most in order to get this fixed :)

KristofferC:
The current strategy is just sneakily making the slowness look like it is the fault of the package depending on CSV. It is significantly increasing the cost of having CSV as a dependency. I really think this is the wrong way of doing it.

quinnj:
Fair enough.

PR #353, opened 2018-11-14
Add more helpful error message on invalid column access for csv rows.…

### LINKED ISSUE #352: New user: ArgumentError: 'nothing' should not be printed
I'm just getting started with Julia and trying to use CSV.jl to read a tab-delimited file.

I'm trying to iterate over the rows, but anything I do with them results in "ArgumentError: `nothing` should not be printed; use `show`, `repr`, or custom output instead."

I have no idea where "nothing" would be entering into the data, as all rows and columns are populated in the source file.  And unless I misunderstand the interface, there's no other way to view the data in the structure to inspect what is happening? Any way to pick out a row or column or entry?

quinnj:
… Fixes #352

### DIFF
--- a/src/tables.jl
+++ b/src/tables.jl
@@ -136,9 +136,12 @@ const FUNCTIONMAP = Dict(
     CatStr => C_NULL,
 )
 
+@noinline badcolumnerror(name) = "`$name` is not a valid column name"
+
 @inline function Base.getproperty(row::Row, name::Symbol)
     f = getfield(row, 1)
     i = findfirst(x->x===name, f.names)
+    i === nothing && throw(ArgumentError(badcolumnerror(name)))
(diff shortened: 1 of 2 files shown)


src/tables.jl
```diff
@@ -136,9 +136,12 @@ const FUNCTIONMAP = Dict(
     CatStr => C_NULL,
 )
 
+@noinline badcolumnerror(name) = "`$name` is not a valid column name"
+
 @inline function Base.getproperty(row::Row, name::Symbol)
     f = getfield(row, 1)
     i = findfirst(x->x===name, f.names)
+    i === nothing && throw(ArgumentError(badcolumnerror(name)))
```
nalimilan: I thought typically you'd also put `throw` in the `@noinline` function?

PR #372, opened 2019-01-11
When ignorerepeated is true, we need to make sure we ignore delimiter…

quinnj:
…s after newlines, this requires a new method in Parsers.jl that allows just checking for and consuming delimiters


jmkuhn:
This works well for my test files.

PR #376, opened 2019-01-14
Introduce `recorddelim` to `CSV.write`

### LINKED ISSUE #375: Custom newline character
For `CSV.write` it would be good to have a custom "end of record" character similar to how we can replace `,` with something else.

In our particular case a client has requested CSV with Windows-style endlines, `"\r\n"`, at the end of each record. While we can post-process the file with `sed`, it would seem reasonable to have another keyword argument to the `CSV.write` function.

andyferris:
Users can specify the delimeter for terminating records as a new keyword argument `recorddelim`. The default remains `'\n'`.

Resolves #375.

If the approach seems reasonable I can add tests and docs.

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -114,17 +114,18 @@ function write(file::Union{String, IO}, itr; kwargs...)
     return write(sch, rows, file; kwargs...)
 end
 
-function printheader(io, header, delim, oq, cq, e, df)
+function printheader(io, header, delim, recorddelim, oq, cq, e, df)
     cols = length(header)
     for (col, nm) in enumerate(header)
         bufferedwrite(io, string(nm), nothing) && bufferedescape(io, delim, oq, cq, e)
-        Base.write(io, ifelse(col == cols, UInt8('\n'), delim))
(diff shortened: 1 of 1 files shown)


src/write.jl
```diff
@@ -154,6 +155,7 @@ end
 # handle unknown schema case
 function write(::Nothing, rows, file::Union{String, IO};
     delim::Union{Char, String}=',',
+    recorddelim::Union{Char, String}='\n',
```
quinnj: I'd rather just call this `newline`

andyferris: Happy to oblige... but technically the record seperator doesn't have to be a line seperator at all.

For example, our literal matrices might have field delimiter `' '` and record delimiter `;` and no newlines.

nalimilan: Maybe `newrow`?

nalimilan:
A bit off-topic, but it would be nice to accept both `\n` and `\r\n` by default, without using a `String` to avoid a performance hit (maybe a tuple of `Char`)?

quinnj:
Implemented in https://github.com/JuliaData/CSV.jl/pull/406

Issue #378, opened 2019-01-21
Support parsing to PooledArrays


nalimilan:
It's been requested several times that we support parsing strings to PooledArrays. That would be more appropriate than CategoricalArray in many situations, and it would be much more efficient (especially for grouping) than `Vector{String}`.

In terms of implementation, it should be quite similar to CategoricalArrays, but there's a difficulty. While we can simply use `CategoricalValue` to carry the information about the parsed value *and* about the type of the column to allocate, there's no equivalent for PooledArrays (which is the whole point of having two packages). I can see two approachs:
- Define a custom `PooledString <: AbstractString` type which would be almost identical to `CategoricalString`, but would be used to indicate that a `PooledArray` should be allocated to store the result. This is my preferred approach.
- Don't use a wrapper when streaming `CSV.File`, but find a way to indicate to `CSV.read` which columns should be `PooledArray`. This would have the drawback that it would be impossible to use when streaming.

matthieugomez:
Thanks for opening this @nalimilan!

quinnj:
Implemented in #395

PR #384, opened 2019-02-10
WIP: Add PooledArray support

nalimilan:
Add a `pool` argument similar to `categorical` to store strings as `PooledArray{String}`.
Like for `CategoricalString`, this is handled via a `PooledString <: AbstractString` type storing a reference to the pool, which implements `Tables.allocatecolumn` in order to be stored as a `PooledArray`.

Fixes https://github.com/JuliaData/CSV.jl/issues/378.

This works but two things need to be discussed. The first one is that even if `PooledString` holds a reference to a pool, we don't take advantage of it at all since `Tables.allocatecolumn` creates a `PooledArray` with a different pool. So setting each entry requires doing a standard dict lookup. Therefore the only use of `PooledString` is to indicate that we want a `PooledArray`. `CategoricalString` has the same issue. I can see two solutions to this:
- Pass the first `PooledString` element (rather than just the type) to `Tables.allocatecolumn` (with a fallback to passing the type). That way the returned `PooledArray` can use the same `Dict` as a pool, and the optimized `setindex!` method defined in this PR can be used to just copy the integer reference (currently this method isn't used e.g. by `CSV.File(...) |> DataFrame`). The same strategy can be used for `CategoricalString`.
- Only store a reference to the string as a `Tuple{Ptr{UInt8}, Int}` or `WeakRefString` in `PooledString`, and leave it to `setindex!` to perform the dict lookup when assigning to the `PooledArray`. I'm not sure whether that's possible WRT the guarantees on the lifetime of the buffer. This approach cannot be used for `CategoricalString`.

The second issue is more minor: do we want `PooledString` to actually be an `AbstractString`, in which case we need to define the methods included in pooledstring.jl. I'd say yes since it's not a lot of code, and it could allow making `pool=true` or `pool=0.25` the default without changing the behavior of parsed strings for users. If we want to keep it I'll copy tests from CategoricalArrays.


quinnj:
Implemented in #395

PR #394, opened 2019-03-26
Relax type constraints on filenames

oxinabox:
To allow this to work with FilePathBase file types.
Which no longer subtype `AbstractString` (https://github.com/rofinn/FilePathsBase.jl/pull/22)

This change basically says tht if you don't give it a `IO` object,
then whatever you do give it must have an `open`  method.


quinnj:
Closing in favor of https://github.com/JuliaData/CSV.jl/pull/405

PR #395, opened 2019-03-27
Overhaul

quinnj:
Ok, here's a rundown of things going on here, I apologize for not chunking this up more, but sometimes a code bomb is a code bomb.

Alright, here's a list of things going on here:
 * Removing old deprecations (this is mostly just deprecated.jl going away + tests)
 * Finishing #384 by adding PooledArray support (though currently untested, but planned before merging)
 * Removal of transforms.jl in favor of the more official code in Tabes.jl
 * Deprecation of CSV.validate; CSV.read now handles all misshapen files and will warn when there are an unexpected # of columns

 The rest is involves a big overhaul of Tables.columns(f::CSV.File). The goal in overhauling is tri-fold:
 1) Improve performance
 2) Avoid dynamic compilation per file
 3) Improve "typing"

 I'm pleased to say we now do very well w/ all three with the changes here. Performance is at least on-par w/ other available solutions (in Julia, R, python, etc.). We now avoid dynamic compilation of any size file (previously it was only for smallish files). And finally, we now
 get "perfect typing" of csv files. Yes, you read that correctly. And no, we don't need to restart parsing at all. Whatever values are encountered while parsing, the resulting array is effectively "promoted", losslessly, to the correct final type, all in one pass over the data.

 Things left to do before merging:
 * Add lots more tests (see PooledArray comment above)
 * Figure out the "streaming" or "iteration" situation: I'd still like to support iterating a File, line by line, but we need to have more restrictions on what types are allowed for columns, since supporting CategoricalArray/PooledArray become prohibitive/dangerous when streaming
 * Clean up the rest of the existing tests to make sure they work in our new world

If anyone feels brave enough to review, I'd really love the eyes: cc @nalimilan @bkamins @omus or anyone else....


davidanthoff:
Cool! Ping me when you think this is ready to be run through https://github.com/davidanthoff/csv-comparison. I can also just do a test run and post results here if that would help.

src/tables.jl
```diff
@@ -1,189 +1,424 @@
...
+                levels!(pool, sort(levels(pool)))
+                A = CategoricalArray{String, 1}(x.refs, pool)
+            end
+        end
+    elseif x.typecode === STRING
+        A = StringArray{x.escapestrings ? E : String, 1}(buf, x.offsets, x.lengths)
+    elseif x.typecode === (STRING | MISSING)
+        A = StringArray{Union{x.escapestrings ? E : String, Missing}, 1}(buf, x.offsets, x.lengths)
+    elseif x.typecode === EMPTY
+        A = Missing[]
```
nalimilan: Should this be `Union{}`?

src/tables.jl
```diff
@@ -1,189 +1,424 @@
...
+    refs::Vector{UInt32}
+    pool::Dict{String, UInt32}
+    poolm::Dict{Union{String, Missing}, UInt32}
+
+    Array(len) = new(Vector{UInt64}(undef, len), Vector{UInt32}(undef, len), EMPTY, false, 0)
+end
+
+# "materialize" a CSV.Array as a real Array/CategoricalArray/PooledArray/StringArray
+# and resize the resulting array according to the # of rows we actually parsed
+function vec(buf, x::Array, rows, E, pool, categorical)
```
nalimilan: It's quite confusing to have a type named `Array` since you can't tell which type this refers to (same below). Maybe use a different name?

src/CSV.jl
```diff
@@ -1,45 +1,46 @@
...
+    Tables.columns(File(fullpath; kwargs...))
+    return
 end
 
-function Base.showerror(io::IO, e::Error)
-    println(io, "CSV.Error on row=$(e.row), column=$(e.col):")
-    showerror(io, e.error)
-end
+import Tables.transform
+export transform
```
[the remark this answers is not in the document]
quinnj: We already export it, so it'd be breaking if we stopped. But maybe we should make that change anyway, since it's a pretty common name and could easily collide out in the wild.

quinnj:
Ok, made some good progress here after taking a week or so detour into some fun advanced performance stuff (that will come later). For now, this is definitely an improvement for the package.

In this latest round of updates, I also added the ability to auto-detect the delimiter for the file, which was pretty easy.

My remaining TODO items here are:
  * add tests for PooledArray
  * add tests for other new functionality (auto-detected delims, some of the perfect column typing code paths)
  * update docs & code comments
  * one more round of performance improvements

I'm hoping to wrap all of this up in the next couple of days.

bkamins:
This is fantastic. I will have a look at the PR when you are done with TODOs (I plan to run it through some sample CSVs as this is the best test for such a complex change 😄).

quinnj:
I've added a Manifeset.toml file on this branch temporarily in case people want to pull this branch and `instantiate` to try it out. There are some changes in WeakRefStrings & Parsers that I'm upstreaming thru the package release process.

quinnj:
Ok, upstream releases have been proposed for [WeakRefStrings](https://github.com/JuliaRegistries/General/pull/248) and [Parsers](https://github.com/JuliaRegistries/General/pull/249), so pending those approvals, I think this is pretty much ready to go. I've updated the Manifest.toml on this branch to reflect the necessary dependencies at the moment, if people want try this branch out (I'd really appreciate it!). I'm going to watch what CI says and see if everything passes (it does for me locally).

quinnj:
Alright, with the new DataFrames.jl release, I think we have a green light forward for merging here. I'm going to do so now, but will hold off on tagging a new release for a few days at least. I think having all these changes on master will hopefully help others test/try things out and we can get a little more feedback on bugs/things to change. I'd also like to take another pass at improving docs/examples before doing a new release.

giordano:
What is supposed to be the replacement of `allowmissing=:none`?  The deprecation warning "`allowmissing` is a deprecated keyword argument" is not 100% clear about this.  It shouldn't be used at all?

nalimilan:
Yes, it's not needed now, as the type will be chosen automatically depending on the presence of missing values. If you want to ensure there are no missing values in your data, you can call `disallowmissing!` on the resulting data frame: it will error if missing values are present.

giordano:
Thanks!

PR #411, opened 2019-05-02
Define a CSV.Column AbstractVector, which is an efficient view into t…

quinnj:
…he underlying csv tape that also reinterprets to the correct data. This provides a convenient, AbstractArray interface while avoiding unnecessary allocations. We also define Base.copy(c::CSV.Column) to materialize a regular Vector, so CSV.read() by default will create a DataFrame w/ CSV.Column columns, while CSV.File() |> DataFrame will fully materialize the results


bkamins:
Thank you for pushing the performance boundary 😄. The thing I was considering is whether there should be a keyword argument that would allow the user to decide what kind of output one wants. The issue is that for small files probably performance does not matter that much and people might prefer a simpler approach. If I understand the design correctly the only change is if we call `DataFrame` or `DataFrame!` at the end of the process - right?

src/utils.jl
```diff
@@ -1,7 +1,52 @@
-const CatStr = CategoricalString{UInt32}
-struct PooledString end
+struct PooledString <: AbstractString
+    ref::UInt64
+    refs::Vector{String}
```
nalimilan: I'd rather call this `pool`.

src/utils.jl
```diff
@@ -1,7 +1,52 @@
...
+end
+
+Base.show(io::IO, x::PooledString) = print(io, '"', escape_string(x.refs[x.ref]), '"')
 export PooledString
 
+Base.get(x::PooledString) = x.refs[x.ref]
```
nalimilan: I don't think this method makes sense here, we use it for categorical values because they wrap another value, but `PooledString` is just a string.

But the `String` and/or `convert` methods should be defined (IIRC the fallbacks are slow since they operate character by character).

src/utils.jl
```diff
@@ -1,7 +1,52 @@
-const CatStr = CategoricalString{UInt32}
-struct PooledString end
+struct PooledString <: AbstractString
+    ref::UInt64
+    refs::Vector{String}
+end
+
+Base.show(io::IO, x::PooledString) = print(io, '"', escape_string(x.refs[x.ref]), '"')
```
nalimilan: Why do you need to escape? I'd just have called `show(io, String(x))`.

nalimilan:
Interesting. A few questions:
- How does this differ from `StringArray`? It it just that it's more general (i.e. it supports numeric and date columns)?
- Isn't there a way to create a standard `PooledArray{String}` without copying refs? I guess that wouldn't make a big difference, but it would be clearer to reuse an existing type (that would also limit recompilation of functions for each column type).
- We'll need to add an efficient constructor in CategoricalArrays to ensure that if people use `pool=true` we directly reuse the refs instead of rehashing everything. Otherwise removing the `categorical` argument will be problematic.

src/utils.jl
quinnj: Sounds good

src/utils.jl
quinnj: Yeah, I was just kind of blindly copying the CategoricalString code; I'll go thru and clean it up.

quinnj:
@bkamins, yes, we could perhaps just include the same `copycols` keyword argument for `CSV.read`, so users don't have to use `CSV.File() |> DataFrame` instead.

@nalimilan, `CSV.Column` is more like a generalized "view" type instead of `StringArray`. It uses the underlying `CSV.File` and `tape::Vector{UInt64}` as "data" and computes the right strides to subindex + reinterpret the data.

bkamins:
Thinking of it,  `copycols` is a good choice as at least when someone learns this pattern once it is consistent across the packages. Thank you.

src/iteration.jl
```diff
@@ -6,58 +6,57 @@ struct Row
...
 @inline function getcell(f::File, ::Type{T}, col::Int, row::Int) where {T}
-    indexoffset = ((f.cols * (row - 1) * 2) + (col - 1) * 2) + 1
-    offlen = f.tape[indexoffset]
+    indexoffset = ((getfield(f, :cols) * (row - 1) * 2) + (col - 1) * 2) + 1
+    offlen = getfield(f, :tape)[indexoffset]
     missingvalue(offlen) && return missing
     return getvalue(Base.nonmissingtype(T), f, indexoffset, offlen, col)
 end
 
 getvalue(::Type{Int64}, f, indexoffset, offlen, col) = int64(f.tape[indexoffset + 1])
```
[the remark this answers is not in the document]
quinnj: Great idea; just pushed accessors

quinnj:
Ok, let's give this a whirl.

src/CSV.jl
```diff
@@ -146,6 +150,10 @@ function File(source;
     (types !== nothing && any(x->!isconcretetype(x) && !(x isa Union), types isa AbstractDict ? values(types) : types)) && throw(ArgumentError("Non-concrete types passed in `types` keyword argument, please provide concrete types for columns: $types"))
     delim !== nothing && ((delim isa Char && iscntrl(delim) && delim != '\t') || (delim isa String && any(iscntrl, delim) && !all(==('\t'), delim))) && throw(ArgumentError("invalid delim argument = '$(escape_string(string(delim)))', must be a non-control character or string without control characters"))
     allowmissing !== nothing && @warn "`allowmissing` is a deprecated keyword argument"
+    if categorical !== false
```
nalimilan: This isn't a real deprecation, as the resulting columns won't be `CategoricalArray`s. How about tagging a release without this PR but with the deprecation first to leave a little time for people to adapt?

src/tables.jl
```diff
@@ -1,225 +1,113 @@
...
-        A = Column(Vector{Union{DateTime, Missing}}(undef, rows))
-    elseif T === BOOL
-        A = Column(Vector{Bool}(undef, rows))
-    elseif T === (BOOL | MISSING)
-        A = Column(Vector{Union{Bool, Missing}}(undef, rows))
+Base.size(c::Column) = (length(c.r),)
+Base.IndexStyle(::Type{<:Column}) = Base.IndexLinear()
+function Base.copy(c::Column{T}) where {T}
+    len = length(c)
+    A = Vector{T}(undef, len)
```
nalimilan: Wouldn't it make sense to allocate a `PooledArray` for pooled columns?

Issue #443, opened 2019-05-24
another mmap error on windows


jaakkor2:
This looks a bit like closed https://github.com/JuliaData/CSV.jl/issues/424 , a bit like https://discourse.julialang.org/t/csv-mmap-error-when-parsing-large-file/24555 , but involves lines to be skipped in the beginning of the file
```julia
hh = true ## additional garbage in the beginning of file
using CSV, DataFrames
cols = 1000
for rows in (10, 100, 1000)
    filename = string("test_", rows, "x", cols, ".csv")
    println("Writing ", filename)
    open(filename, "w") do io
    	hh && write(io, "foo\nbar\n")
    end
    df = DataFrame(rand(rows, cols))
    CSV.write(filename, df, writeheader=true, append=true)
	a = CSV.read(filename, header=1+2*hh, datarow=2+2*hh)
end
```

Output looks like
```
Writing test_10x1000.csv
Writing test_100x1000.csv
Writing test_1000x1000.csv
ERROR: could not create file mapping: The operation completed successfully.
Stacktrace:
 [1] error(::String) at .\error.jl:33
 [2] #mmap#1(::Bool, ::Bool, ::Function, ::Mmap.Anonymous, ::Type{Array{UInt64,1}}, ::Tuple{Int64}, ::Int64) at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\Mmap\src\Mmap.jl:218
 [3] #mmap#14 at .\none:0 [inlined]
 [4] mmap at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\Mmap\src\Mmap.jl:251 [inlined]
 [5] file(::String, ::Int64, ::Bool, ::Int64, ::Nothing, ::Int64, ::Int64, ::Bool, ::Nothing, ::Bool, ::Array{String,1}, ::String, ::Nothing, ::Bool, ::Char, ::Nothing, ::Nothing, ::Char, ::Nothing, ::UInt8, ::Nothing, ::Nothing, ::Nothing, ::Nothing, ::Dict{Int8,Int8}, ::Bool, ::Float64, ::Bool, ::Bool, ::Bool, ::Bool, ::Nothing) at C:\Users\j\.julia\packages\CSV\IwqOm\src\CSV.jl:278
[...]

PR #444, opened 2019-05-25
Add a CSV.View type that ignores any type information while parsing a…

quinnj:
…nd provides the same interface as CSV.File

cc: @GregPlowman

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -613,6 +613,7 @@ end
 
 include("tables.jl")
 include("iteration.jl")
+include("views.jl")
 include("write.jl")
 
 """
--- a/src/views.jl
+++ b/src/views.jl
@@ -1,0 +1,292 @@
+struct View
+    name::String
+    names::Vector{Symbol}
+    rows::Int64
+    cols::Int64
+    e::UInt8
+    buf::Vector{UInt8}
+    tapes::Vector{Vector{UInt64}}
+    anymissing::Vector{Bool}
+end
+
+getname(f::View) = getfield(f, :name)
+getnames(f::View) = getfield(f, :names)
+getrows(f::View) = getfield(f, :rows)
+getcols(f::View) = getfield(f, :cols)
+gete(f::View) = getfield(f, :e)
+function getrefs(f::View, col)
+    @inbounds r = getfield(f, :refs)[col]
+    return r
+end
+getbuf(f::View) = getfield(f, :buf)
+function gettape(f::View, col)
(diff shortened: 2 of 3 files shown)


nalimilan:
Can you explain what's the advantage of this?

quinnj:
Parsing an entire file via `CSV.View` is about 10-20% faster, while using about half the memory footprint. So it should alleviate problems that users sometimes run into on windows like #443. What they get back is essentially an object w/ the same interface as `CSV.File`, but each column is assumed to be String type.

Given how easy it was to implement this, and for the amount of code duplication, I'm leaning towards not merging this due to the relatively meager performance gain. I'm also working on the idea of a `CSV.Rows` type to iterate rows w/ a minimal memory footprint, also assuming `String` type for each column that I think provides a similar benefit with essentially no memory footprint.

GregPlowman:
Thanks for looking at this.

> Parsing an entire file via CSV.View is about 10-20% faster, while using about half the memory footprint. So it should alleviate problems that users sometimes run into on windows like #443. What they get back is essentially an object w/ the same interface as CSV.File, but each column is assumed to be String type.
> 
> Given how easy it was to implement this, and for the amount of code duplication, I'm leaning towards not merging this due to the relatively meager performance gain. I'm also working on the idea of a CSV.Rows type to iterate rows w/ a minimal memory footprint, also assuming String type for each column that I think provides a similar benefit with essentially no memory footprint.

Well for me it's not necessarily the "performance gain" that is most important (but faster is better), but whether arbitrarily large files can be processed or not. Halving the memory footprint is significant, but there will always be a file twice as large that will have the same issue of running out of memory.

An approach that scales indefinitely with file size would seem useful. Files could be processed row-by-row, with fixed memory requirements. Of course there could be optimizations such as buffering in chunks etc.

nalimilan:
OK, thanks. I agree duplicating all of that code isn't necessarily worth it: as @GregPlowman says, what would be very useful is a way to avoid allocating a copy at all (which the old implementation did, but with issues regarding type detection).

quinnj:
> An approach that scales indefinitely with file size would seem useful. Files could be processed row-by-row, with fixed memory requirements. Of course there could be optimizations such as buffering in chunks etc.

Yeah, that's why I'm favoring #445 over this PR, because it can handle arbitrarily large files w/ a fixed-size buffer. There are still ways we can optimize the memory use of CSV.File, even incorporating some of the ideas here, so I think I'll close this PR for now.

PR #466, opened 2019-07-05
Fix parsing for date/times with custom dateformat

ibacher:
Without the try / catch blocks, `detect()` fails when trying to parse something
similar to a date, but not a valid date (e.g., a US 5 + 4 zip code).

Adding the try / catch simply makes this test behave identically to the test
when not using a custom date format.

I've also added a test to `basics.jl` for this behaviour.


quinnj:
Thanks @ibacher !

PR #470, opened 2019-07-10
Fix an issue where we weren't matching pooled strings correctly becau…

quinnj:
…se we stored the *escaped* string in our pool, but were comparing the *unescaped* to see if we've seen that value before. Luckily, the fix is pretty simple, if the string needs escaping, we need to escape it before comparing with our pool values. Definitely one of those bugs that took way longer to track down and figure out what exactly was going wrong than to actually fix it. Fixes https://github.com/JuliaData/DataFrames.jl/issues/1880


bkamins:
Great. Thank you!

PR #473, opened 2019-07-13
CSV.File(::Cmd) support

aaronmichaelsilberstein:
Makes CSV.read behave a bit more like Base.read, by allowing to read commands directly.


test/basics.jl
```diff
@@ -340,4 +340,21 @@ df = CSV.read(IOBuffer("a,b,c\n1,2,3\n\n"), ignoreemptylines=true)
 df = CSV.read(IOBuffer("zip\n11111-1111\n"), dateformat = "y-m-dTH:M:S.s")
 @test size(df) == (1, 1)
 
+# Supporting commands across multiple platforms cribbed from julia/test/spawn.jl
+catcmd = `cat`
+havebb = false
+if Sys.iswindows()
+    busybox = download("https://frippery.org/files/busybox/busybox.exe", joinpath(tempdir(), "busybox.exe"))
```
quinnj: According to [this SO question](https://superuser.com/questions/434870/what-is-the-windows-equivalent-of-the-unix-command-cat), it seems like we could just call `type $file`?

aaronmichaelsilberstein: I thought so too.  But commit 2bde38b doesn't pass regression...

src/utils.jl
```diff
@@ -152,6 +152,8 @@ end
 function getsource(source, use_mmap)
     if source isa Vector{UInt8}
         return source
+    elseif source isa Cmd
+        return Base.read(source)
```
quinnj: Hmmmm, this makes me a bit worried, if you happened to do something like `CSV.read(`gunzip really_large_csv.csv.gz`)`, we might run out of memory, but I'm also not sure what other option we would have. I guess this is probably fine.

quinnj:
Sorry for the delay here; I think this looks good; can you rebase and then I can merge?

quinnj:
Thanks!

PR #479, opened 2019-08-07
small cleanup of delimiter validation

ExpandingMan:
I've exanded the set of valid delimiters so that we support all control characters except for `'\n'`, `'\r'` and `'\0'`.  There doesn't seem to be a particularly good reason to forbid the others (I've recently encountered a situation in which I was asked to use `Char(0x01)` and this seems to work ok).  I've also cleaned things up a bit so that there is a common function for checking whether a delimiter is valid, and for throwing an error if it is not.  Finally, I've added a check to `CSV.write` so that we forbid writing CSV's that we refuse to read (previously you could even write one with `'\n'` as a delimiter).

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -92,7 +92,7 @@ Checks whether the argument is valid for use as a data source, otherwise throws
 an error.
 """
 function checkvalidsource(source)
-    !isa(source, IO) && !isa(source, Vector{UInt8}) && !isfile(source) &&
+    !isa(source, IO) && !isa(source, Vector{UInt8}) && !isa(source, Cmd) && !isfile(source) &&
         throw(ArgumentError("\"$source\" is not a valid file"))
 end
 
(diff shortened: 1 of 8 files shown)


src/CSV.jl
```diff
@@ -61,6 +61,41 @@ const EMPTY_TYPEMAP = Dict{TypeCode, TypeCode}()
...
+end
+
+"""
+    checkvalidsource(source)
+
+Checks whether the argument is valid for use as a data source, otherwise throws
+an error.
+"""
+function checkvalidsource(source)
+    !isa(source, IO) && !isa(source, Vector{UInt8}) && !isfile(source) &&
```
quinnj: Can you update this to account for https://github.com/JuliaData/CSV.jl/pull/473 that I just merged?

ExpandingMan: Done

Issue #432, opened 2019-05-10
Memory usage for large files


bkamins:
Following https://github.com/h2oai/db-benchmark/pull/85#issuecomment-491340029 we have an issue in H2O benchmarks for large files.

@quinnj - do you think it is related to a conversion using `DataFrame`? The error seems to indicate that we have a problem earlier.

nalimilan:
So IIUC they are using a VM with 125GB of RAM and trying to load a 50GB CSV file. That should be enough to hold both the file contents and the tape in RAM.

quinnj:
Is there a way to "retry" the benchmark to see if it's better on the latest 0.5.3 release? The most recent release should be a little better on memory pressure, but I'd like to know if we need to dig in further.

bkamins:
If you want to reproduce the benchmark data exactly [here](https://github.com/h2oai/db-benchmark/blob/master/README.md) are the instructions in **Reproduce** section (but probably producing large files with structure similar to described [here](https://h2oai.github.io/db-benchmark/) will be good enough :)) - I never tried to reproduce the steps described in the official guide.

nalimilan:
Apparently they ran the benchmarks on July 29th (so with recent package versions), and they still run out of memory.

quinnj:
@bkamins, do you know if there's a way to provide the types explicitly for the h20 benchmark? If so, that would drastically improve the memory footprint with #510 merged.

nalimilan:
I think we can basically make a PR to apply any changes we want at https://github.com/h2oai/db-benchmark/blob/master/juliadf/groupby-juliadf.jl#L23. They shouldn't object since the benchmark is about grouping, not CSV reading.

nalimilan:
Oh, and do you think that without changing that line, https://github.com/JuliaData/CSV.jl/pull/510 may improve memory use enough that it may now work? We don't really need reading the data to be fast (it's not included in the benchmark), it just has to work.

bkamins:
Yes - in general in H2O benchmark, we should change to the option that is most memory efficient for the larges data set case (it does not have to be the fastest). And any PR we would submit there should be accepted.

bkamins:
H2O benchmark still produces "out of memory" due to CSV.File. See https://github.com/h2oai/db-benchmark/blob/7e178c1d2fb9102c8b12ac201f883981254a9df6/benchplot-dict.R#L171.
for the detailed list of offending tests

@quinnj - is anything still fixable here? Thank you!

[...]

PR #481, opened 2019-08-20
Utilize multithreading while parsing

quinnj:
Ok, this is in decent shape now. I've seen some performance improvements ranging from modest to excellent.

There are a few more TODOs here, including:
* in the `findrowstarts!`, we're currently not accounting for the case where we might drop in mid-quoted-cell and incorrectly "find" a newline that's quoted
* ~~What if nthreads == 1 and `threaded=true`; does that currently work?~~ works.
* ~~what to do about Julia versions < 1.3~~ add appropriate `@static` VERSION check, and warn if you're on an unsupported version + pass `threaded=true`
* a little more type stability work; it's a bit tricky w/ a few closures and the `Threads.@spawn`, but I think it's just a matter of cleaning up variable names/usage

But I'd appreciate people trying this out if they're so inclined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ os:
 
 julia:
   - 1.0
-  - 1.1
+  - 1.2
+  - 1.3
   - nightly
 
 env:
@@ -27,7 +28,7 @@ after_success:
 jobs:
   include:
     - stage: "Documentation"
-      julia: 1.1
+      julia: 1.2
       os: linux
       script:
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate(); Pkg.build("CSV")'
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -305,10 +305,10 @@ function file(source,
     end
     debug && println("column names detected: $names")
     debug && println("byte position of data computed at: $datapos")
-    if threaded === nothing && VERSION >= v"1.3" && Threads.nthreads() > 1 && !transpose && limit == typemax(Int64) && footerskip == 0 && rowsguess >= 1_000
+    if threaded === nothing && VERSION >= v"1.3-DEV" && Threads.nthreads() > 1 && !transpose && limit == typemax(Int64) && footerskip == 0 && rowsguess >= 1_000
         threaded = true
     elseif threaded === true
-        if VERSION < v"1.3"
+        if VERSION < v"1.3-DEV"
             @warn "incompatible julia version for `threaded=true`: $VERSION, requires >= v\"1.3\", setting `threaded=false`"
             threaded = false
         end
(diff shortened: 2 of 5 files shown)


quinnj:
Ok, the latest commit cleans up performance and type stability; as a quick sample, with the Fannie Mae `Performance_2016Q1.txt` [file](https://www.fanniemae.com/portal/funding-the-market/data/loan-performance-data.html) of 13M rows, I'm seeing:
```julia
julia> @time f = CSV.File(file, pool=false);
  4.857984 seconds (877.92 k allocations: 46.119 MiB, 0.24% gc time)

julia> @time f = CSV.File(file, threaded=false, pool=false);
 10.294354 seconds (411.69 k allocations: 22.460 MiB)
```
vs. `fread`
```R
> system.time(f = fread(file))
Error in system.time(f = fread(file)) : unused argument (f = fread(file))
> system.time(fread(file))
|--------------------------------------------------|
|==================================================|
|--------------------------------------------------|
|==================================================|
   user  system elapsed
 11.204   1.113  12.452
```
I think my `fread` should be using threads here, but I can't quite tell, if anyone can help me out that'd be great.

Drvi:
I think `fread` has a `nThread` kwarg which you can try to vary. This looks awesome, btw:)

xiaodaigh:
I got 

7.67s and 4.0 in 1st and 2nd run for 

```julia
@time f = CSV.File(file, pool=false);
```

and I got "1.430s elapsed (4.080s cpu)" for
```r
file="c:/data/Performance_All/Performance_2016Q1.txt"
pt = proc.time()
a = data.table::fread(file, sep="|", header = FALSE)
timetaken(pt)
```
This is run on my 6 core (12 threads) i7 with 64G with 2 SAMSUNG 850 in RAID0.

I also ran the 2003Q3 which is the largest file there with 148m rows. And it took **42s** for data.table, but I got an error with Julia 1.3.0-rc1.0 and notice the warning? Which says threading didn't work.

```julia
 @time a = CSV.File(file, pool = false, threaded = true);
┌ Warning: incompatible julia version for `threaded=true`: 1.3.0-rc1.0, requires >= v"1.3", setting `threaded=false`
└ @ CSV C:\Users\RTX2080\.julia\packages\CSV\CYBRQ\src\CSV.jl:312
ERROR: SystemError: mmap: The operation completed successfully.
```


> Ok, the latest commit cleans up performance and type stability; as a quick sample, with the Fannie Mae `Performance_2016Q1.txt` [file](https://www.fanniemae.com/portal/funding-the-market/data/loan-performance-data.html) of 13M rows,

It's only 6.5m right? The 2016Q1 Performance data.

xiaodaigh:
> I think my `fread` should be using threads here,

It is using threads by default IMSMR.

quinnj:
I just downloaded the 2016Q1 data yesterday, and I'm seeing 13377198 rows in both Julia & R dataframes, the file is 1.17 GB uncompressed.

Here are my timings again:
```julia
julia> using CSV
[ Info: Precompiling CSV [336ed68f-0bac-5ca0-87d4-7b16caf5d00b]

julia> file = "/Users/jacobquinn/Downloads/2016Q1/Performance_2016Q1.txt"
"/Users/jacobquinn/Downloads/2016Q1/Performance_2016Q1.txt"

julia> f = nothing; GC.gc(); GC.gc(); @time f = CSV.File(file, header=false);
 11.828778 seconds (46.75 M allocations: 1.578 GiB, 3.07% gc time)

julia> f = nothing; GC.gc(); GC.gc(); @time f = CSV.File(file, header=false);
  9.417777 seconds (6.10 k allocations: 569.563 KiB)

julia> f = nothing; GC.gc(); GC.gc(); @time f = CSV.File(file, header=false);
  6.008193 seconds (6.10 k allocations: 569.563 KiB)
```
vs data.table
```R
> library(data.table)
data.table 1.12.2 using 8 threads (see ?getDTthreads).  Latest news: r-datatable.com
(failed reverse-i-search)`file = ':
> file = "/Users/jacobquinn/Downloads/2016Q1/Performance_2016Q1.txt"
> pt = proc.time()
> a = data.table::fread(file, sep="|", header = FALSE)
|--------------------------------------------------|
|==================================================|
Warning message:
In require_bit64() :
  Some columns are type 'integer64' but package bit64 is not installed. Those columns will print as strange looking floating point data. There is no need to reload the data. Simply install.packages('bit64') to obtain the integer64 print method and print the data again.
> timetaken(pt)
[1] "5.859s elapsed (18.5s cpu)"
```
So that seems pretty competitive. It's kinda tricky to benchmark these because they're so memory-intensive, so you get a lot of system variables interacting and depending on what's already "hot" can affect things.

@xiaodaigh, I also tried the 2003Q3 and ran into memory issues. We have an issue, #432, open that I need to work on; currently CSV.jl is very liberal w/ memory use, so we just need to be better about that and I think things will be better for larger files.

I'm finishing up my last TODO on this right now and then I'll push and hopefully we'll get clean CI. If anybody else wants to review or has comments, please post.

quinnj:
Alright let's rock n' roll. I just pushed the last commit where we make sure threads are ok parsing by correctly adjusting to the start of a new row in the case our "hop" ends up in a quoted cell. Very hard to test w/ an actual file since it depends on how many threads are on a machine, so for now I've just included some unit tests for `CSV.findrowstarts!`. It's annoying that it takes so much code to account for an extremely rare use-case, but since it's possible, we'll account for it properly.

Once CI passes, I'm planning on merging.

xiaodaigh:
Maybe Fannie updates the data. I downloaded 1 year ago

quinnj:
Oops, one more update here; I forgot I was going to clean up the checks and docs. Here's the logic: `threaded=nothing` is the default, in that case, we'll check if `transpose=true` and if you've passed anything for `limit` and if you're on the right Julia version; if all those checks pass, we'll check that the guessed # of rows * number of columns (= # of cells) is >= 5_000. I did some quick benchmarks, and that's about the number when using threads starts to give us a speed up on a mixed-type dataset. If you pass `threaded=true` and the checks don't pass, we'll set `threaded = false` and warn you that something was wrong (too old Julia, `transpose=true`, etc.), otherwise, that let's the user use threads on smaller datasets if they're so inclined. `threaded=false` will always avoid the multithreading branch.

zjpi:
> Maybe Fannie updates the data. I downloaded 1 year ago

Maybe this is why the file changed 
https://www.fanniemae.com/portal/funding-the-market/data/loan-performance-dataset-loans-less-30-years-092217.html

PR #491, opened 2019-08-30
updated with gzip and zip examples.

xiaodaigh:


bkamins:
@xiaodaigh maybe it would be good to add code that closes the opened streams in the examples?

PR #504, opened 2019-10-01
add CategoricalArrays.jl version 0.7 support

### LINKED ISSUE #502: Update compatibility with CategoricalArrays.jl
CSV.jl requires CategoricalArrays.jl to be 0.5 or 0.6 only.

@nalimilan - given the new release when this should be synced? Now?

bkamins:
Fixes #502

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CSV"
 uuid = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 authors = ["Jacob Quinn <quinn.jacobd@gmail.com>"]
-version = "0.5.12"
+version = "0.5.13"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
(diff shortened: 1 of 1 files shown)


bkamins:
@quinnj can you please have a look at this (I expect that some packages in ML ecosystem will depend on CategoricalArrays.jl 0.7 soon and then we will start getting allowed version conflicts). Thank you!

bkamins:
Thank you! @quinnj Are you planning to add something else to CSV.jl before tagging the release?

quinnj:
I've been working on a refactoring of the internals for #432, but if you want to do a quick patch release for this change, feel free.

bkamins:
Thank you for looking into #432!

I would do a patch release (as now we have a conflict with released CategoricalArrays.jl 0.7, and it introduces a breaking functionality).

Just, for safety, please confirm that #497 can be safely included in this patch release (I have looked through it and I think it can, but maybe there is something breaking I did not spot).

quinnj:
Yes, that's fine.

Issue #507, opened 2019-10-03
Speed up getproperty(r::Row2, nm::Symbol)?


JockLawrie:
Hi there,

I've recently given [Schemata.jl](https://github.com/JockLawrie/Schemata.jl) a makeover and now I'm working on performance improvements. I am testing with files around 50M records, so small improvements add up.

Having profiled these tests, one opportunity for a speed up is a Dict lookup for `getproperty(r::Row2, nm::Symbol)` instead of using `findfirst`.

Here's a not-very-rigorous MWE:
```julia
a = [Symbol("x$(i)") for i = 1:9]
d = Dict(Symbol("x$(i)") => i for i = 1:9)

@time findfirst(==(:x1), a)
@time d[:x1]
```
On my machine the dict lookup takes about half the time of the array lookup, and one less allocation. Although both times are small, in my case I am executing `getproperty` more than 100M times per test case, so the difference adds up.

Would you be open to including in `Rows` a `Dict` that maps column name to its index in `Rows.names`? Each `Row2` would then inherit this lookup, so it need only be created once.

quinnj:
This is implemented in https://github.com/JuliaData/CSV.jl/pull/510

Issue #501, opened 2019-09-17
`propertynames` does not return column names on a `CSV.Row2`


CameronBieganek:
```
julia> using CSV

julia> io = IOBuffer("x,y\n1,3\n2,4");

julia> rows = CSV.Rows(io);

julia> row1 = first(rows);

julia> propertynames(row1)
(:names, :tape, :buf, :e)
```

The following line from the `CSV.Rows` doc leads me to believe that `propertynames(row1)` should return a tuple or array with the names of the columns.

>Each row object supports propertynames, getproperty, and getindex to access individual row values.

(`propertynames` does return an array of column names when called on a `CSV.Row`.)

quinnj:
Fixed in https://github.com/JuliaData/CSV.jl/pull/510

Issue #500, opened 2019-09-17
CSV.Row2 supports getindex, but CSV.Row does not


ianfiske:
Just checking if there is a reason why we don't want to have `getindex` for `CSV.Row`.  It makes some operations a lot easier.  It's already there for `CSV.Row2` https://github.com/JuliaData/CSV.jl/blob/f046da7d3c42fb435cee8de6947ed314260d4a54/src/rows.jl#L224.

quinnj:
Fixed in https://github.com/JuliaData/CSV.jl/pull/510

Issue #488, opened 2019-08-25
memory mapping failed: Cannot allocate memory


aplavin:
Reading many and/or large CSV files starts giving `ERROR: SystemError: memory mapping failed: Cannot allocate memory` from time to time, even with `use_mmap = false`. Can't see any relevant options to control mem-mapping further.

A simple example:
```julia
# create CSV file
CSV.write("tmp.csv", DataFrame(a=1:1000, b=1:1000))

# read it many times
[CSV.read("tmp.csv", use_mmap=false) for i in 1:1000];
# running the previous line several times reliably gives the error for me
```

Julia 1.2.0, CSV 0.5.11.

quinnj:
I can't reproduce this in https://github.com/JuliaData/CSV.jl/pull/510, even when doing `[CSV.read("tmp.csv", use_mmap=false) for i in 1:10000];`, i.e. 10,000 times instead of just 1,000. The memory efficiency has been improved in that branch, which would explain the improvement.

quinnj:
Please re-open/comment if you still see an issue

FredSoftwares:
I confirm that the problem still exists with Julia 1.4.1 and it is related to multithread and probably hyperthreading [see this post](https://discourse.julialang.org/t/memory-mapping-failed-when-reading-many-csvs/27979/12)

Test done with :
(@v1.4) pkg> status
Status `~/.julia/environments/v1.4/Project.toml`
  [336ed68f] CSV v0.6.2
  [861a8166] Combinatorics v1.0.1
  [a93c6f00] DataFrames v0.21.0

PR #510, opened 2019-10-07
Refactor internals to allow better memory efficiency

### LINKED ISSUE #508: bug with `comment` flag in `read`
contents of `hmm.csv`:
```
# a comment                                                                                 
# no header...                                                                              
"Yes","Medium rare"                                                                         
"Yes","Medium"                                                                              
"Yes","Medium"                                                                              
"Yes","Medium rare"                                                                         
"No",                                                                                       
"Yes","Rare"    
```

```
using CSV
df_steak = CSV.read("steak/hmm.csv", comment="#", header=[:eats_steak, :how_cooked])
```

gives:
```
thread = 1 warning: only found 1 / 2 columns on data row: 2. Filling remaining columns with `missing`
thread = 1 warning: only found 1 / 2 columns on data row: 3. Filling remaining columns with `missing`
thread = 1 warning: only found 1 / 2 columns on data row: 5. Filling remaining columns with `missing`
thread = 1 warning: only found 1 / 2 columns on data row: 6. Filling remaining columns with `missing`
```
and incorrectly reads in the data. I think it has to do with the `comment` since if I delete the comments in the file it reads in fine.

CSV v0.5.12

quinnj:
Also fixes #508, #507, #501, #500, #488, #484, #432, and #424.

To aid in reviewing such a large PR, I'm going to add some more detailed comments around the specific code changes, and people can chime in with thoughts/concerns where appropriate.


src/CSV.jl
```diff
@@ -20,35 +20,40 @@ end
...
     names::Vector{Symbol}
     types::Vector{Type}
     rows::Int64
     cols::Int64
-    e::UInt8
-    categorical::Bool
-    refs::Vector{Vector{String}}
-    buf::Vector{UInt8}
-    tapes::Vector{Vector{UInt64}}
+    columns::Vector{AbstractVector}
```
quinnj: The change here is moving from storing the raw tapes + additional info to just storing the final, `Column` structures directly. That moves a bit of work from `getproperty` on `CSV.File` previously to `CSV.File` construction.

quinnj: The main advantage of this change is it removes a bunch of duplicated logic that existed between tables.jl and iteration.jl before, since row iterating on `CSV.File` now just makes a lazy `Row` wrapper that indexes the already-created `CSV.Column`s.

src/CSV.jl
```diff
@@ -96,6 +101,17 @@ function checkvalidsource(source)
         throw(ArgumentError("\"$source\" is not a valid file"))
 end
 
+function allocate(rowsguess, ncols, typecodes)
+    tapes = Vector{UInt64}[Mmap.mmap(Vector{UInt64}, usermissing(typecodes[i]) ? 0 : rowsguess) for i = 1:ncols]
```
quinnj: Changes here include:
  * if a user passes `Missing` as a column type, then we avoid allocating any tape/poslen vectors at all; any value found in the cell while parsing will still be parsed, but no value/position/length will be stored anywhere. The `CSV.Column{Missing}` just checks that `getindex` is in bounds and then returns `missing`. This can help a lot for large files that have entire columns of null values (though you have to pass `Missing` as the column type manually). It also provides a way we could do really efficient "baked in" column selection while parsing; we would just set the non-selected columns as having column type `Missing` and they essentially get skipped
  * The other change here is that if the user provides the column type for any other kind of column, we don't allocate the `poslen` vector. The `poslen` vector for a column tracks the position and length of each cell, in case we need to promote the column to a string at any point later on while parsing. For user-provided types, the user is essentially declaring the column type should be X, so we don't need to track this additional info (which costs memory). Previously, the column `tape` stored the raw parsed values and poslen values in the same array; separating them into `tape` and `poslen` allows us to avoid the `poslen` when it's not needed.

src/CSV.jl
```diff
@@ -364,141 +380,138 @@ function file(source,
...
+            else
+                finalrefs[i] = nothing
             end
         end
+    else
+        fill!(finalrefs, nothing)
     end
-    return File(getname(source), names, finaltypes, rows, ncols, eq, categorical, finalrefs, buf, tapes)
+    if threaded === true
+        columns = AbstractVector[ApplyArray(vcat, (Column{_eltype(finaltypes[i]), finaltypes[i]}(tapes[j][i], rows[j], eq, categorical, finalrefs[i], buf, finaltypes[i] >: Int64 ? uint64(INT_SENTINELS[intsentinel.x]) : sentinelvalue(Base.nonmissingtype(finaltypes[i]))) for j = 1:Threads.nthreads())...) for i = 1:ncols]
```
quinnj: For the multithreaded case, each thread parses an equal chunk of the file. Previously, we were then allocating a final, huge tape and copying each thread's chunk into the final tape.

Here, we're using the quite-well-optimized `LazyArray.Vcat` type to lazily `vcat` the threads' tapes into a final "column". So essentially in the non-threaded case, you get back a `Vector{Column}`, and in the multithreaded case, you get back `Vector{LazyArrays.Vcat{Column}`. I've tried to do a pretty thorough benchmarking/stress-testing on `LazyArray.Vcat` and I've been really impressed with the optimizations they've implemented for broadcasting, reductions, etc. 

I also feel like we've already conditioned users to know that in the `copycols=false` case, you're getting some kind of optimized "view" `AbstractVector`, and to use `copycols=true` if they need a final, full `Vector`.

src/CSV.jl
```diff
@@ -364,141 +380,138 @@ function file(source,
...
+    else
+        columns = AbstractVector[Column{_eltype(finaltypes[i]), finaltypes[i]}(tapes[i], rows, eq, categorical, finalrefs[i], buf, finaltypes[i] >: Int64 ? uint64(INT_SENTINELS[intsentinel.x]) : sentinelvalue(Base.nonmissingtype(finaltypes[i]))) for i = 1:ncols]
+    end
+    lookup = Dict(k => v for (k, v) in zip(names, columns))
```
quinnj: I've seen a couple comments and we had at least one issue opened about the cost of `getproperty` on `CSV.Row` and `CSV.Row2`, so throwing a lookup Dict into `CSV.File` and `CSV.Rows` seems like not a high price to pay for faster property access later.

src/CSV.jl
```diff
@@ -364,141 +380,138 @@ function file(source,
...
                         end
                     end
-                    recodes[i, col] = recode
-                    rngs[i, col] = rng
-                end
-            end
-        end
-    end
-    @sync for j = 1:N
-@static if VERSION >= v"1.3-DEV"
```
quinnj: By using `LazyArrays`, it also made me realize we can clean up the ref-array syncing code in the multithreaded case. Instead of a two-step, complicated process, we now just do the single loop above and recode a pooled columns refs if needed.

src/CSV.jl
```diff
@@ -561,89 +577,96 @@ function parsetape(::Val{transpose}, ignoreemptylines, ncols, typemap, tapes, ta
...
                 end
             end
-            tapeidx += 2
             pos > len && break
-            if tapeidx + 1 > tapelen
-                debug && reallocatetape()
-                oldtapes = tapes
-                newtapelen = ceil(Int64, tapelen * 1.5)
-                newtapes = Vector{UInt64}[Mmap.mmap(Vector{UInt64}, newtapelen) for i = 1:ncols]
+            # if our initial row estimate was too few, we need to reallocate our tapes/poslens to read the rest of the file
```
quinnj: This should be a much smarter algorithm for re-allocating column tapes if we've underestimated the # of rows to parse. Before, we were just blindly doing `tapelen * 1.5`, which could be disastrous on really large files. We now look at how many bytes are left in the file, how many rows we've parsed vs. how many rows we've parsed, and make a smarter guess at how many rows are left in the file. We then allocate the new tapes one at a time, freeing the previous column's old tape before allocating the next column's new tape.

src/CSV.jl
```diff
@@ -704,34 +758,47 @@ end
...
         else
             y, code, vpos, vlen, tlen = Parsers.xparse(Float64, buf, pos, len, options)
             if code > 0
-                @inbounds tape[tapeidx + 1] = uint64(y)
-                @inbounds typecodes[col] = FLOAT | (missingtype(T) ? MISSING : EMPTY)
+                # recode past Int64 values
+                intsent = uint64(INT_SENTINELS[intsentinel.x])
+                for i = 1:(row - 1)
+                    @inbounds z = tape[i]
+                    @inbounds tape[i] = ifelse(z === intsent, sentinelvalue(Float64), uint64(Float64(int64(z))))
```
quinnj: In the past, we didn't go back and recode already-parsed Int64 values to Float64, but stored a bit to say that it was originally a `Int64` or not. We now don't keep the extra `poslen` vectors around after parsing (since when we finish, we then know the perfect types for each column and can free that memory up).

Now, if we have been parsing an `Int64` column that needs to be promoted to `Float64`, we just go recode the past values, which makes less work in `CSV.Column` indexing and one less thing to keep track of.

src/iteration.jl
```diff
@@ -20,50 +20,58 @@ end
...
+@noinline badcolumnerror(name) = throw(ArgumentError("`$name` is not a valid column name"))
 
-# internal method for getting a cell value
-@inline function getcell(f::File, ::Type{T}, col::Int, row::Int) where {T}
-    ind = metaind(row)
-    @inbounds offlen = gettape(f, col)[ind]
-    missingvalue(offlen) && return missing
-    return getvalue(Base.nonmissingtype(T), f, ind, offlen, col)
+@inline function Base.getproperty(row::Row{Int}, name::Symbol)
+    column = getcolumn(getfile(row), name)
```
quinnj: This is where we get to remove a lot of duplicated logic from tables.jl and just index the `CSV.Column` for our `CSV.Row`.

src/rows.jl
```diff
@@ -228,5 +234,24 @@ Base.size(r::Row2) = (length(getnames(r)),)
...
-Base.getproperty(r::Row2, nm::Symbol) = getindex(r, findfirst(==(nm), getnames(r)))
 
+Base.propertynames(r::Row2) = getnames(r)
+
+function Base.getproperty(r::Row2, nm::Symbol)
+    @inbounds x = r[getlookup(r)[nm]]
+    return x
+end
+
+@inline Base.@propagate_inbounds function Parsers.parse(::Type{T}, r::Row2, i::Int) where {T}
```
quinnj: These two functions now allow an even more efficient way to parse values when iterating `CSV.Rows`, like:
```julia
for row in CSV.Rows(file)
    int = Parsers.parse(Int64, row, :column1)
end
```
which will avoid the intermediate `String` allocation when doing `row.column1`.

src/utils.jl
```diff
@@ -97,15 +99,15 @@ gettypecodes(x::Dict{TypeCode, TypeCode}) = x
 # bit patterns for missing value, int value, escaped string, position and len in tape parsing
 const MISSING_BIT = 0x8000000000000000
 missingvalue(x::UInt64) = (x & MISSING_BIT) == MISSING_BIT
+sentinelvalue(::Type{Float64}) = Core.bitcast(UInt64, NaN) | MISSING_BIT
```
quinnj: One wrinkle I didn't anticipate from separating our previous `tape` into `tape` + `poslen` is how to encode `missing` values. Before we always had a `poslen` value around where we could steal a bit, but now we don't keep `poslen` around after parsing. It turns out that for all but `Int64`, we can find an empty bit pattern to use as a sentinel for `missing` (we use an obscure `NaN` bit pattern for `Float64`, and `Core.bitcast(UInt64, typemin(Int64))` for all other types).

I wrestled with what to do for `Int64`, since every bit pattern is a valid `Int64`. I first considered `typemin(Int64)`, but worried that it would turn out to actually be more frequent of a value encountered in the wild than a bunch of random `Int64`s (I'm thinking about manually encoded sentinels where they use `typemin(Int64)` themselves to mean something special). The approach here has 4 random Int64 values, positive and negative, and if we encounter one, we'll use the next one and recode the old sentinels. I think statistically we're in the clear here, but it is a bit unfortunate that this is so much uglier than other types.

src/utils.jl
```diff
@@ -334,31 +336,48 @@ Base.IndexStyle(::Type{ReversedBuf}) = Base.IndexLinear()
...
+        return new{T}(A)
     end
 end
 
 Base.size(a::AtomicVector) = size(a.A)
 Base.IndexStyle(::Type{A}) where {A <: AtomicVector} = Base.IndexLinear()
 
-Base.getindex(a::AtomicVector, i::Int) = a.A[i][]
-Base.setindex!(a::AtomicVector, x, i::Int) = setindex!(a.A[i], x)
+for typ in Base.Threads.atomictypes
```
quinnj: This updates our `AtomicVector` type to be more efficient. Instead of storing a `Threads.Atomic{T}` for each element and having extra pointer indirections for each load/store, we generate the llvm intrinsic instructions ourselves for `getindex`/`setindex!` for each type.

Long term it'd be great to have a ThreadedDataStructures.jl package or something where this can live, but for now, this does what we want/need for multithreaded parsing, and it's pretty simple.

For reference, I basically copied the `llvmcall`s from Base and used [CudaNative.jl](https://github.com/JuliaGPU/CUDAnative.jl/blob/master/src/device/cuda/atomics.jl) and tips from @vchuravy/@maleadt.

src/utils.jl
```diff
@@ -334,31 +336,48 @@ Base.IndexStyle(::Type{ReversedBuf}) = Base.IndexLinear()
...
+                 """, $typ, Tuple{Ptr{$typ}}, pointer(x.A, i))
+    @eval Base.setindex!(x::AtomicVector{$typ}, v::$typ, i::Int) =
+        Base.llvmcall($"""
+                 %ptr = inttoptr i$(Base.Threads.WORD_SIZE) %0 to $lt*
+                 store atomic $lt %1, $lt* %ptr release, align $(Base.Threads.gc_alignment(typ))
+                 ret void
+                 """, Cvoid, Tuple{Ptr{$typ}, $typ}, pointer(x.A, i), v)
+end
+
+function unset!(A::Vector, i::Int, row, x)
```
quinnj: We're being a bit tricky here, but it's in the name of memory efficiency. First we `finalize` our mmapped array, which calls `munmap` underneath, then we call `jl_arrayunset`, which deletes the `Vector{UInt64}` Julia structure. This will mostly be called when columns get promoted from some type to `String`, in which case we don't need both `tape` and `poslen`, so we get rid of `poslen` to avoid extra allocations.

src/utils.jl
nalimilan: You mean that an error is thrown if the 4 values are encountered in the data? I guess that's OK, but OTC it would be a real problem if a missing value was silently generated instead of the actual value.

quinnj: Hmmm, actually we don't throw an error if we encounter all 4, but we probably should do a fatal error, or perhaps allow the user to manually pass a sentinel to use for Int.

nalimilan: I'd throw an error for now and see whether that works. Providing a sentinel manually is really not user-friendly.

BTW, rather than random values, we can probably find good values by hand. For example, `typemax(Int)-1` is a really good candidate. OTC, anything close to 0 should be avoided.

bkamins: Why do we choose 4 sentinels. My intuition is that `INT_SENTINEL` could be a `Ref`. We initialize it somehow (here we can discuss). If we hit a collision (in practice I think it will be extremely unlikely) then we just carefully generate a new value to store in this `Ref` so that we are sure it was never encountered yet. In this way we will have no limie of 4 values, and moreover the code should be simpler (we do not need to keep track of the counter).

bkamins: Of course I might be missing something - I have just glanced over the code so please comment.

bkamins: And in this case this value should be local to a given column. An alternative would be to keep using the vector as you do and just grow it as needed (then it can be shared across in the single julia process).

quinnj: Here's how my thinking here has evolved, to catch everyone up:

* Ok, need to figure out a sentinel for `Int64`, R uses `typemin(Int64)`, so we could maybe do that
* But being `typemin(Int64)` might actually be too common if people hand encode things, so maybe `typemin(Int64) + 1`
* That's fine, but what if we happen to encounter that value while parsing? We could potentially just add 1 each time we encounter our sentinel, but that would be bad if they happen to have a sequence of values that correspond w/ our sequence of sentinels
* A BIG issue with changing the sentinel though, is that each time, in addition to going back and recoding past stored sentinel values to a new sentinel value, you're also worried about every past parsed value being equal to the new sentinel; I guess maybe that cost is worth simplifying things; i.e. if we happen to encounter our int sentinel, we pick a new Int64 sentinel randomly (not totally random, probably more like `rand(typemin(Int64):div(typemin(Int64), 10))`), and then, in addition to going back and recoding past sentinels, we also check that every parsed Int64 doesn't equal our new sentinel. Ultimately, that's not _that_ expensive of an operation in addition to having to recode our sentinel being very statistically unlikely.

quinnj: > (making sure to lock this vector in case multithreading is used).

So..........it turns out, I somehow completely didn't consider the multithreaded case in all of this and it's all obviously wrong (the recoding, bumping sentinels, and not just for Int types, but other types are affected as well). This probably explains why we still see issues in #495 and #484. I think I have an idea of how to adjust all this for multithreaded, so I'll take a stab at it.

bkamins: Thank you - please let me know when you are done so we can come back to this excellent PR.

quinnj: Ok, just pushed some new commits. The changes are these:

* Instead of using a single `AtomicVector` that each thread accesses/updates, we have a per-thread types vector
* So each thread will start "fresh" when parsing, detect values, and update/promote types as it's chunk of the file is parsed
* After all threads have finished, and similar to what we were already doing for pooled columns, we now "sync" the types detected from each thread and adjust if needed; e.g. if one thread parsed all Ints, it will have an Int column type, but another thread for the same column may have encountered a String value and promoted it. In our "sync" phase, we promote the types from each thread for each column, so that column's type would be promoted to String and we would recode the Int column values in that thread.
* The recoding/promoting of the "sync" phase isn't too bad/complex, and should be fairly rare, so I'm not too worried about any kind of serious performance hit.
* For the Int64 sentinel story, we now allocate an Int64 sentinel per column per thread, and do a similar "sync" afterwards, to ensure each column is encoded to a single sentinel value. It's a pain and a lot of code for how statistically unlikely it will be to ever hit that case, but whatever, we handle it and I just need to find a way to test it. Basically the logic is: do the per-thread parsing, afterwards, we set our `intsentinels` equal to the first thread's `intsentinels`, then check if any other column in any other thread ended up w/ a different int sentinel than the starting one (already pretty unlikely). If so, we then bump that column's int sentinel to the one that thread ended up (since we know it's valid for that thread). We then have a pass if a column's int sentinel isn't the starting one, to check if the new int sentinel is equal to any parsed value in any other thread. If so, we bump the int sentinel again and re-check all the threads until we find a sentinel that isn't present in the parsed value of any thread. We then do the actual recoding of each thread's sentinel values to the new, unique/good int sentinel for that column. Whew......

quinnj:
Ok, update here: I pushed some more fixes that I think gets this into "correct" territory. I still want to run some more performance tests on large files and add some tests for some of the new code paths, but I'd appreciate if anyone has any comments here or more simply just tries this branch out on a few files and lets me know if they see anything (`] add CSV#jq/mem3` to get this branch).

src/utils.jl
bkamins: I think that in order to test the `Int64` sentinel story we have to generate the data that contains exactly the sentinels that you are expecting to use (I have not dug in the code yet, but I assume that you are generating them in a way that can be made reproducible e.g. by setting appropriate seeds for PRNG).

src/CSV.jl
```diff
@@ -96,14 +101,30 @@ function checkvalidsource(source)
...
+        end
+    end
+    return tapes, poslens
+end
+
 """
     CSV.File(source; kwargs...) => CSV.File
 
-Read a csv input (a filename given as a String or FilePaths.jl type, or any other IO source), returning a `CSV.File` object.
+Read a UTF8 csv input (a filename given as a String or FilePaths.jl type, or any other IO source), returning a `CSV.File` object.
```
nalimilan:
```suggestion
Read a UTF-8 CSV input (a filename given as a String or FilePaths.jl type, or any other IO source), returning a `CSV.File` object.
```

src/CSV.jl
```diff
@@ -96,14 +101,30 @@ function checkvalidsource(source)
...
 
-Opens the file and uses passed arguments to detect the number of columns and column types.
+Opens the file and uses passed arguments to detect the number of columns and column types, unless column types are provided
+manually via the `types` keyword argument. Note that passing column types manually can increase performance and reduce the
+memory use for each column type provided (column types can be given as a `Vector` for all columns, or specified per column via 
+name or index in a `Dict`). For text encodings other than UTF8, see the [StringEncodings.jl](https://github.com/JuliaStrings/StringEncodings.jl)
```
nalimilan:
```suggestion
name or index in a `Dict`). For text encodings other than UTF-8, see the [StringEncodings.jl](https://github.com/JuliaStrings/StringEncodings.jl)
```

src/CSV.jl
```diff
@@ -116,7 +137,7 @@ end
 By supporting the Tables.jl interface, a `CSV.File` can also be a table input to any other table sink function. Like:
 
 ```julia
-# materialize a csv file as a DataFrame
+# materialize a csv file as a DataFrame, allowing DataFrames to take ownership of the CSV.File columns
```
nalimilan: "Take ownership" isn't very explicit without more explanation. Maybe just "without copying columns"?

src/CSV.jl
```diff
@@ -364,141 +385,228 @@ function file(source,
...
 end # @static if VERSION >= v"1.3-DEV"
     end
-    rows = sum(rowsv)
-    tapes = Vector{UInt64}[Mmap.mmap(Vector{UInt64}, rows * 2) for i = 1:ncols]
+    intsentinels = perthreadintsentinels[1]
+    anyintrecode = false
+    # promote typecodes from each thread
+    for col = 1:ncols
+        for i = 1:N
+            @inbounds typecodes[col] = promote_typecode(typecodes[col], perthreadtypecodes[i][col])
```
nalimilan: Why apply `@inbounds` only to this line?

src/CSV.jl
quinnj: Updated

src/CSV.jl
quinnj: Applied `@inbounds` to the other line

xiaodaigh:
Just tested latest master and CSV.jl is basically on par with data.table on my 16G RAM laptop on thsi 5GB CSV https://www.kaggle.com/kernels?sortBy=dateRun&group=profile&pageSize=20

PR #511, opened 2019-10-07
Ensure FilePaths work

morris25:
The CSV.write docstring claims that it should work with FIlePaths
> Write a Tables.jl interface input (https://github.com/JuliaData/Tables.jl) to a csv file, given as an IO argument or String/FilePaths.jl type representing the file name to write to.

However, when I try to write to a Path I get 
```
CSV.write(mktmpdir() / "test.txt", DataFrame(:a=>[1,2,3]))
ERROR: MethodError: no method matching with(::getfield(CSV, Symbol("##48#49")){Bool,Tables.Schema{(:a,),Tuple{Int64}},Tables.RowIterator{NamedTuple{(:a,),Tuple{Array{Int64,1}}}},CSV.Options{UInt8,UInt8,Nothing,Tuple{}},Tuple{Symbol},Int64,Int64,Array{UInt8,1}}, ::PosixPath, ::Bool)
Closest candidates are:
  with(::Function, ::Union{DevNull, Pipe, PipeEndpoint, TTY}, ::Any) at /Users/sam/.julia/dev/CSV/src/write.jl:136
  with(::Function, ::IO, ::Any) at /Users/sam/.julia/dev/CSV/src/write.jl:130
  with(::Function, ::String, ::Any) at /Users/sam/.julia/dev/CSV/src/write.jl:140
Stacktrace:
 [1] #write#47(::Bool, ::Bool, ::Array{String,1}, ::Function, ::Tables.Schema{(:a,),Tuple{Int64}}, ::Tables.RowIterator{NamedTuple{(:a,),Tuple{Array{Int64,1}}}}, ::PosixPath, ::CSV.Options{UInt8,UInt8,Nothing,Tuple{}}) at /Users/sam/.julia/dev/CSV/src/write.jl:73
 [2] write(::Tables.Schema{(:a,),Tuple{Int64}}, ::Tables.RowIterator{NamedTuple{(:a,),Tuple{Array{Int64,1}}}}, ::PosixPath, ::CSV.Options{UInt8,UInt8,Nothing,Tuple{}}) at /Users/sam/.julia/dev/CSV/src/write.jl:68
 [3] #write#46(::Char, ::Char, ::Nothing, ::Nothing, ::Char, ::Char, ::Char, ::Nothing, ::Bool, ::String, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(CSV.write), ::PosixPath, ::DataFrame) at /Users/sam/.julia/dev/CSV/src/write.jl:60
 [4] write(::PosixPath, ::DataFrame) at /Users/sam/.julia/dev/CSV/src/write.jl:53
 [5] top-level scope at none:0
```

These changes should ensure CSV read/write functions work for any AbstractPath.


Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.5.13"
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+FilePaths = "8fc22ac5-c921-52a6-82fd-178b2807b824"
```
nickrobinson251: Would `FilePathsBase` be sufficient? To avoid depending on MacroTools and URIParser

mattBrzezinski:
Looks good to me, only nice to have would be a test case with your example from the CR description:

> CSV.write(mktmpdir() / "test.txt", DataFrame(:a=>[1,2,3]))

src/utils.jl
```diff
@@ -174,6 +174,8 @@ function getsource(source, use_mmap)
         return source
     elseif source isa Cmd
         return Base.read(source)
+    elseif source isa AbstractPath
+        return Base.read(Base.open(source))
```
quinnj: Hmmm, I don't love this because we're not matching the behavior when `source` is a String, which I think should be consistent. When `source` is a String and we have `use_mmap`, then we `Mmap.mmap`, otherwise, we make our own mmapped vector and copy the file contents into it. I'd rather rearrange to ensure strings and abstractpath sources are treated the same.

oxinabox:
The previous PR to add FilePaths support was 
https://github.com/JuliaData/CSV.jl/pull/394

It probably should have added FilePaths as a Test dependendcy, so that we would have been better at not breaking it

src/utils.jl
```diff
@@ -189,6 +191,7 @@ end
 
 getname(buf::Vector{UInt8}) = "<raw buffer>"
 getname(cmd::Cmd) = string(cmd)
+getname(path::AbstractPath) = string(path)
 getname(str) = String(str)
```
oxinabox: THis line was added in #394 to support FilePaths types.
Then FilePaths change from using `String` to usinmg `string` to get back to a plain string path type.
(which I still think is a dubious choice, but 🤷‍♂)

I take it is is not surficient anymore to  just do 
```suggestion
getname(str) = string(str)
```

So that means this line is probably not used anymiore
and so we can just:
```suggestion
```

src/utils.jl
oxinabox: Note that some FilePaths can be `mmap`s directly, in particular `SystemPath`s
other cannot, like `AWS3.S3Path`s
But I guess that can be fairly easily handled by treating `SystemPath`s like strings,
and all other `AbstractPaths` like `Cmd`s etc.

https://github.com/rofinn/FilePathsBase.jl/issues/51

src/write.jl
```diff
@@ -136,7 +136,7 @@ function with(f::Function, io::Union{Base.TTY, Base.Pipe, Base.PipeEndpoint, Bas
     f(io)
 end
 
-function with(f::Function, file::String, append)
+function with(f::Function, file::Union{String, AbstractPath}, append)
```
nalimilan: BTW, shouldn't this use `AbstractString`?

src/utils.jl
morris25: `SystemPath`s don't directly support `mmap` and have to be opened first. We're also stuck using open on non-System Paths since that guarantees us the right type coming out of `read` (`S3Path` right now returns a `String`). 

Also why are we checking `slurp(source isa IO ? source : open(String(source)))` after the `elseif !isa(source, IO)` in this function?

Manifest.toml
```diff
@@ -0,0 +1,239 @@
+[[Base64]]
```
quinnj: Please don't include a Manifest file

morris25: Sorry, I'm not sure how I managed to inadvertently add that.

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.5.13"
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"
```
nickrobinson251: Should there be a [compat] entry for FilePathsBase?

test/write.jl
```diff
@@ -197,4 +202,9 @@ using CSV, Dates, WeakRefStrings, CategoricalArrays, Tables
     end
 
     @test read(rd, String) == "col1,col2,col3\n1,4,7\n2,5,8\n3,6,9\n"
+
+    mktmpdir() do tmp
```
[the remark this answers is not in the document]
rofinn: `mktmdir()` wasn't deprecated, it's just an alias for `mktempdir(SystemPath)` now.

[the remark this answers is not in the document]
nickrobinson251: ok

src/utils.jl
```diff
@@ -178,12 +178,13 @@ end
...
         return source
-    elseif source isa Cmd
-        return Base.read(source)
-    elseif use_mmap && !isa(source, IO)
-        return Mmap.mmap(source)
+    elseif source isa Cmd || (source isa AbstractPath && !isa(source, SystemPath))
+        return Base.read(open(source))
     elseif !isa(source, IO)
-        m = Mmap.mmap(source)
+        m = Mmap.mmap(source isa AbstractPath ? open(source) : source)
```
oxinabox: I think this is effectively the same as
```suggestion
        m = open(Mmap.mmap, source)
```
since all `Mmap.mmap(source::String)` does is basically `open(Mmap.mmap, source)` anyway

oxinabox:
My alternative dispatch based `getsource`:
First two cold be combined via a `Union`


```
getsource(source::Cmd, ::Any) = Base.read(open(source))
getsource(source::AbstractPath, ::Any) = Base.read(open(source))
getsource(source::Vector{UInt8}, ::Any) = source
getsource(source::IO, ::Any) = slurp(source)
getsource(source::SystemPath, use_mmap) = getsource(string(source), use_mmap)
function getsource(source, use_mmap)
    m = Mmap.mmap(source)
    use_mmap && return m
        
    m2 = Mmap.mmap(Vector{UInt8}, length(m))
    copyto!(m2, 1, m, 1, length(m))
    finalize(m)
    return m2
end
```

quinnj:
Also, you'll probably need to rebase off current master head/tip to get windows CI fixes for travis.

quinnj:
Bump; is there anything else @morris25 want to do on this? Any concerns @mattBrzezinski or @oxinabox? If this can be rebased, then I think CI should be clean and I'm happy to merge and do a new release.

morris25:
That should be everything now! Sorry I didn't get back to this sooner

src/utils.jl
```diff
@@ -175,27 +175,25 @@ function slurp(source)
     return final
 end
 
+getsource(source::Vector{UInt8}, ::Any) = source
```
mattBrzezinski: I was just typing this solution up as you pushed your commit.

Thank you for making this 100% cleaner and easier to read!! 🚀 💯

Issue #512, opened 2019-10-07
jq/mem3 getting errors; due to using LazyArrays@0.10.x


xiaodaigh:
Loading the Fannie Mae 2004Q3 performance data gives the below.

```julia
julia> @time df = CSV.read("c:/data/perf/Performance_2004Q3.txt");
ERROR: MethodError: no method matching ndims(::LazyArrays.Applied{LazyArrays.DefaultArrayApplyStyle,typeof(vcat),NTuple{6,CSV.Column{Int64,Int64}}})
Closest candidates are:
  ndims(::Base.Generator) at generator.jl:53
  ndims(::Number) at number.jl:67
  ndims(::Type{#s66} where #s66<:Number) at number.jl:68
  ...
Stacktrace:
 [1] LazyArrays.ApplyArray{Any,N,F,Args} where Args<:Tuple where F where N(::LazyArrays.Applied{LazyArrays.DefaultArrayApplyStyle,typeof(vcat),NTuple{6,CSV.Column{Int64,Int64}}}) at C:\Users\RTX2080\.julia\packages\LazyArrays\HSJVe\src\lazyapplying.jl:133
 [2] LazyArrays.ApplyArray(::LazyArrays.Applied{LazyArrays.DefaultArrayApplyStyle,typeof(vcat),NTuple{6,CSV.Column{Int64,Int64}}}) at C:\Users\RTX2080\.julia\packages\LazyArrays\HSJVe\src\lazyapplying.jl:134
 [3] LazyArrays.ApplyArray(::Function, ::CSV.Column{Int64,Int64}, ::Vararg{CSV.Column{Int64,Int64},N} where N) at C:\Users\RTX2080\.julia\packages\LazyArrays\HSJVe\src\lazyapplying.jl:138
 [4] file(::String, ::Int64, ::Bool, ::Int64, ::Nothing, ::Int64, ::Int64, ::Bool, ::Nothing, ::Bool, ::Bool, ::Array{String,1}, ::String, ::Nothing, ::Bool, ::Char, ::Nothing, ::Nothing, ::Char, ::Nothing, ::UInt8, ::Nothing, ::Nothing, ::Nothing, ::Nothing, ::Dict{Int8,Int8}, ::Bool, ::Float64, ::Bool, ::Bool, ::Nothing, ::Bool, ::Bool, ::Nothing) at C:\Users\RTX2080\.julia\packages\CSV\2hvaW\src\CSV.jl:433
[...]

PR #513, opened 2019-10-08
adding LazyArrays.jl requirements

xiaodaigh:
See #512


Project.toml
```diff
@@ -23,6 +23,7 @@ Parsers = "0.3"
 PooledArrays = "0.5"
 Tables = "0.1,0.2"
 WeakRefStrings = "0.5,0.6"
+LazyArrays = "^0.12.0"
```
quinnj: this should just be `LazyArrays = "0.12"`

xiaodaigh: updated

quinnj: Thanks!

PR #517, opened 2019-10-16
Remove redunctance check in getsource

oxinabox:
Noticed in #511 
am I right?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -189,7 +189,8 @@ function getsource(source, use_mmap)
         finalize(m)
         return m2
     else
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -189,7 +189,8 @@ function getsource(source, use_mmap)
         finalize(m)
         return m2
     else
-        return slurp(source isa IO ? source : open(String(source)))
+        assert(isa(source, IO))
```
quinnj: Well, in current master, `FilePath`s are handled in this branch, so this will definitely break `FilePath` support. Better just finish up #511

PR #525, opened 2019-11-07
Make CSV.File subtype AbstractVector and cleanup iteration

quinnj:
I'd love to get feedback from @nalimilan, @bkamins, and @andyferris on this. Changes include:

* Adding a `threaded` type parameter to `CSV.File`; this is convenient for dispatching for non-threaded vs. threaded files, and avoids at least one additional dynamic dispatch while not adding a ton of overhead (i.e. a single boolean type parameter)
* Makes the subtyping `struct File{threaded} <: AbstractVector{Row{threaded}}` official
* Adds a `getindex` method on `CSV.File`

I will note that iterating and accessing row values on threaded files currently has a performance cost vs. non-threaded case due to the LazyArray tuple of columns not being fully encoded in the `CSV.File` object. That feels like drastically too high a cost (essentially encoding all columns/types in `CSV.File`). I feel like the threaded iteration perf cost is somewhat justified though, since 1) there's no way we're encoding the full column types in `CSV.File`, 2) we have `CSV.Rows` which provides a much more reasonable iteration flow for large, threaded files, and 3) you can pass `threaded=false` and get super fast iteration if you need. A bonus reason is we can always hope/count on compiler magic to make things better/faster.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -37,7 +37,31 @@ _eltype(::Type{Union{PooledString, Missing}}) = Union{String, Missing}
 Base.size(c::Column) = (c.len,)
 Base.IndexStyle(::Type{<:Column}) = Base.IndexLinear()
 
-struct File
+struct Row{threaded}
+    names::Vector{Symbol}
+    columns::Vector{AbstractVector}
+    lookup::Dict{Symbol, AbstractVector}
+    row::Int64
+    array_index::Int64
+    array_i::Int64
+end
+
+getnames(r::Row) = getfield(r, :names)
+getcolumn(r::Row, col::Int) = getfield(r, :columns)[col]
+getcolumn(r::Row, col::Symbol) = getfield(r, :lookup)[col]
+getrow(r::Row) = getfield(r, :row)
+getarrayindex(r::Row) = getfield(r, :array_index)
+getarrayi(r::Row) = getfield(r, :array_i)
+
+Base.propertynames(r::Row) = getnames(r)
+
+function Base.show(io::IO, r::Row)
+    print(io, "CSV.Row($(getrow(r))): ")
+    names = getnames(r)
+    show(IOContext(io, :compact => true), NamedTuple{Tuple(names)}(Tuple(getproperty(r, nm) for nm in names)))
+end
+
+struct File{threaded} <: AbstractVector{Row}
     name::String
     names::Vector{Symbol}
     types::Vector{Type}
@@ -52,6 +76,8 @@ getnames(f::File) = getfield(f, :names)
 gettypes(f::File) = getfield(f, :types)
 getrows(f::File) = getfield(f, :rows)
 getcols(f::File) = getfield(f, :cols)
+getcolumns(f::File) = getfield(f, :columns)
+getlookup(f::File) = getfield(f, :lookup)
 getcolumn(f::File, col::Int) = getfield(f, :columns)[col]
 getcolumn(f::File, col::Symbol) = getfield(f, :lookup)[col]
 
@@ -60,6 +86,11 @@ function Base.show(io::IO, f::File)
     println(io, "Size: $(getrows(f)) x $(getcols(f))")
     show(io, Tables.schema(f))
 end
+
+Base.IndexStyle(::Type{File}) = Base.IndexLinear()
+Base.eltype(f::File{threaded}) where {threaded} = Row{threaded}
+Base.size(f::File) = (getrows(f),)
+Base.Vector{AbstractVector}(f::File) = getcolumns(f)
 
 const EMPTY_POSITIONS = Int64[]
 const EMPTY_TYPEMAP = Dict{TypeCode, TypeCode}()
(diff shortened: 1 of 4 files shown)


src/CSV.jl
```diff
@@ -1028,6 +1058,8 @@ Parses a delimited file into a `DataFrame`. `copycols` determines whether a copy
 """
 read(source; copycols::Bool=false, kwargs...) = DataFrame(CSV.File(source; kwargs...), copycols=copycols)
 
+DataFrames.DataFrame(f::CSV.File; copycols::Bool=false) = DataFrame(getcolumns(f), getnames(f); copycols=copycols)
```
quinnj: This is a slightly unfortunate fallout of `CSV.File <: AbstractVector`, since the fallback `DataFrame` constructor special cases that check and it fails for the case when `CSV.File` has zero rows.

src/iteration.jl
```diff
@@ -2,76 +2,85 @@
...
+    return Row{false}(getnames(f), getcolumns(f), getlookup(f), row, 0, 0)
 end
 
-getfile(r::Row) = getfield(r, :file)
-getrow(r::Row) = getfield(r, :row)
-
-Base.propertynames(r::Row) = getnames(getfile(r))
+@inline Base.@propagate_inbounds function Base.getindex(f::File{true}, row::Int)
+    @boundscheck checkbounds(f, row)
+    c = getcolumn(f, 1)
```
bkamins: what if `f` has `0` columns?

bkamins:
Looks nice. I have left one comment. In general the threaded case is hard to decode with logic how setting `array_index` and `array_i` should work, but I guess you know what you are doing here 😄.

src/iteration.jl
quinnj: If `f` has 0 columns, then it _has_ to have 0 rows, which get checked in the `checkbounds(f, row)` above.

bkamins: right 👍

src/CSV.jl
andyferris: It's unfortunate that you need to define this at all, you mean?

src/iteration.jl
```diff
@@ -2,76 +2,85 @@
 Tables.rowaccess(::Type{<:File}) = true
 Tables.rows(f::File) = f
 
-struct Row{T}
-    file::File
-    row::T
+@inline Base.@propagate_inbounds function Base.getindex(f::File{false}, row::Int)
```
andyferris: `@propagate_inbounds` already implies `@inline`

src/CSV.jl
```diff
@@ -52,6 +76,8 @@ getnames(f::File) = getfield(f, :names)
 gettypes(f::File) = getfield(f, :types)
 getrows(f::File) = getfield(f, :rows)
```
andyferris: I'm wondering if `getnrows` and `getncols` is better (the distinction between `getcols` and `getcolumns` is confusing). Similarly the `File` struct could have field names `nrows` and `ncols`.

src/CSV.jl
```diff
@@ -61,6 +87,10 @@ function Base.show(io::IO, f::File)
     show(io, Tables.schema(f))
 end
 
+Base.IndexStyle(::Type{File}) = Base.IndexLinear()
+Base.eltype(f::File{threaded}) where {threaded} = Row{threaded}
+Base.size(f::File) = (getrows(f),)
+
```
andyferris: I think one usually defines `Base.axes` also?

```
Base.axes(f::File) = (Base.OneTo(getrows(f)),)
```

andyferris: I suppose it would make sense to place the `getindex` implementation here since it's a part of the `AbstractArray` interface, keeping it together.

src/iteration.jl
```diff
@@ -2,76 +2,85 @@
...
-    column = getcolumn(getfile(row), name)
+@inline function Base.getproperty(row::Row{false}, col::Symbol)
+    column = getcolumn(row, col)
     @inbounds x = column[getrow(row)]
     return x
 end
 
-@inline function Base.getindex(row::Row{Int}, i::Int)
-    column = getcolumn(getfile(row), i)
+@inline function Base.getindex(row::Row{false}, col::Int)
```
andyferris: Should this pretend to be a `NamedTuple` at bit more by also allowing indexing by `Symbol`?

src/iteration.jl
```diff
@@ -2,76 +2,85 @@
...
+        i -= n
+    end
+    return Row{true}(getnames(f), getcolumns(f), getlookup(f), row, 1, 1)
+end
 
-function Base.show(io::IO, r::Row)
-    println(io, "CSV.Row($(getrow(r))) of:")
-    show(io, getfile(r))
+# non-threaded file
+@inline function Base.iterate(f::File{false}, st::Int=1)
```
andyferris: This implementation should be free from `AbstractArray`.

src/iteration.jl
quinnj: Oh nice!

src/CSV.jl
quinnj: These aren't meant to be publicly used at all, so I didn't put a ton of thought into the names.

src/iteration.jl
quinnj: Good call.

src/iteration.jl
quinnj: Ah, you're probably right; but I do want to provide an overload for the `threaded` case, because we can be more efficient.

src/CSV.jl
quinnj: That's the default definition, so I think we're good.

andyferris: Hmm... right you are! Though I wonder if `Base` has it backwards - `size` could derive from `axes` without making any assumptions.

src/iteration.jl
andyferris: That's fine; there won't be a method ambiguity.

src/CSV.jl
nalimilan: Do you think something can be improved in DataFrames? Otherwise, doesn't sound too bad.

andyferris: I wouldn’t say it’s too bad, either. But of course, the best interfaces support blind composition... so I guess we could look at the constructors at some point?

nalimilan: Woops, something I missed is that `copycols` should definitely default to `true` here, for consistency with the generic method defined in DataFrames. Also this PR is breaking because of that: people who did `df = DataFrame(CSV.File(...)); append!(df, row)` will now get an error, but previously it worked. I've filed https://github.com/JuliaData/CSV.jl/pull/531.

PR #531, opened 2019-11-17
Default to  `copycols=true` in `DataFrames` constructor

nalimilan:
This is required for consistency with the generic `DataFrames` constructor.
Reverting this avoids introducing a breaking change, as `df = DataFrame(CSV.File(...)); append!(df, row)` used to work and has to keep working.


quinnj:
Thanks; good catch.

PR #562, opened 2020-01-12
Added treatnothingasmissing option

tk3369:
Due to https://github.com/JuliaLang/julia/pull/32148, CSV now writes files with "nothing" in the CSV file.  This is arguably not the right behavior.  

Here's the MWE:
```
julia> CSV.write("/tmp/tom.csv", [(1,2,nothing)], writeheader = false); readlines("/tmp/tom.csv")
1-element Array{String,1}:
 "1,2,nothing"
```

From [Slack discussion](https://julialang.slack.com/archives/C674VR0HH/p1578678980221600), an idea is to allow `nothing` to be treated like `missing` so the user does not need to translate the data to `missing` before writing a CSV file, which is a huge convenient factor.  
This PR adds a new `treatnothingasmissing` option.  Behavior as follows:

1. When it is set to `true`, the writer will treat `nothing` like `missing`.  A positive effect is that we can now print `nothing` as `NA` if `missingstring` is also specified.  
2. When it is set to `false`, the writer will output an empty string for `nothing`, which should be a better default for sane people.

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -10,6 +10,7 @@ Supported keyword arguments include:
 * `openquotechar::Char`: instead of `quotechar`, use `openquotechar` and `closequotechar` to support different starting and ending quote characters
 * `escapechar::Char='"'`: ascii character used to escape quote characters in a text field
 * `missingstring::String=""`: string to print for `missing` values
+* `treatnothingasmissing::Bool=false`:  `nothing` is treated like `missing` or like empty string
 * `dateformat=Dates.default_format(T)`: the date format string to use for printing out `Date` & `DateTime` columns
 * `append=false`: whether to append writing to an existing file/IO, if `true`, it will not write column names by default
 * `writeheader=!append`: whether to write an initial row of delimited column names, not written by default if appending
@@ -30,6 +31,7 @@ mutable struct Options{D, N, DF, M}
     dateformat::DF
     quotestrings::Bool
     missingstring::M
+    treatnothingasmissing::Bool
 end
 
 tup(x::Char) = x % UInt8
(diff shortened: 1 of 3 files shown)


nalimilan:
As noted on Slack, I think it would be better to allow the user to pass an arbitrary function to be called on each cell. Then users would simply do `transform=x -> something(x, missing) `. That should be as efficient and it would avoid adding yet another very specific option (also the name isn't great :-).

tk3369:
I'm unsure if that will be efficient but we can always test.  

I can work on a separate PR for that idea.  Looking at the [writerow function](https://github.com/JuliaData/CSV.jl/blob/master/src/write.jl#L180), we could possibly support a `transform` function that takes two arguments - `col` and `val`, so the user can do different things for different columns.  The default value for the `transform` function could be an identity-like function: `(col, val) -> val`.

@quinnj your thoughts?

src/write.jl
```diff
@@ -185,6 +188,17 @@ function writerow(buf, pos, len, io, sch, row, cols, opts)
...
 end
 
+# Special treatment for `nothing`. It can be treated like `missing`
+# if the `treatnothingasmissing` option is set to `true`.  Otherwise,
+# an empty string is written.
+function writecell(buf, pos, len, io, ::Nothing, opts)
+    if opts.treatnothingasmissing
+        writecell(buf, pos, len, io, missing, opts)
+    else
+        writecell(buf, pos, len, io, "", opts)
```
quinnj: I'm not comfortable with this; I think this branch should throw an informative error, pointing the user to `treatnothingasmissing=true` or using `Tables.transform`. The point of `nothing` in the language is to be a non-useful "null" value that is dealt with as quickly as possible; I think it's an anti-pattern to start giving it behavior.

tk3369: Fair enough :-)

tk3369: @quinnj I have updated code to throw error when `nothing` is encountered.  I also added the column index to the `writecell` function everywhere so we can give the user a little more clue about where the `nothing` was found.  I was thinking about adding the row number too if we like this approach.

@nalimilan I did a separate implementation to allow passing in a cell-based transformer function.  That works too but my performance test shows 2x regression for a 100000x3 data frame of Float64's.  I don't know how to optimize it.  You can find the code at my fork https://github.com/tk3369/CSV.jl/tree/cell-based-transform

Here's the perf test harness:
```
$ cat test/perf_write.jl 
using DataFrames, CSV, BenchmarkTools

function test(n)
    df = DataFrame(x = rand(n), y = rand(n), z = rand(n))
    @benchmark CSV.write("/tmp/tom.csv", $df)
end

for i in 4:7
    rows = 10^i
    println(rows, " rows: ", test(rows))
end
```

Performance result from this PR:
```
10000 rows: Trial(3.081 ms)
100000 rows: Trial(41.997 ms)
1000000 rows: Trial(415.127 ms)
10000000 rows: Trial(5.391 s)
```

Performance result from `cell-based-transform` branch:
```
10000 rows: Trial(6.696 ms)
100000 rows: Trial(78.961 ms)
1000000 rows: Trial(779.919 ms)
10000000 rows: Trial(10.350 s)
```

Performance from version 0.5.22:
```
10000 rows: Trial(2.589 ms)
100000 rows: Trial(40.224 ms)
1000000 rows: Trial(418.302 ms)
10000000 rows: Trial(5.852 s)
```

nalimilan: You have to make `transform::F` with `F<:Union{Function, Nothing}` in the definition of `Options` to avoid dynamic dispatch.

Also you could simplify the code a bit by using `(col, val) -> val` instead of `nothing` as the default: the compiler will probably be able to optimize this out.

tk3369: Submitted PR #564. Your suggestion solved the performance issue.  

We can close this PR after you guys review the other one and agree that is the right direction.

tk3369:
closed in favor of #564

PR #563, opened 2020-01-16
update description of CI specification

bkamins:
I proposed a generic text so that we do not have to update README.md as new stable releases of Julia are shipped (currently we test against 1.3 not 1.1).


bkamins:
@quinnj - if you feel this should be handled differently can you please just close this PR? Thank you!

PR #564, opened 2020-01-19
Cell based transform on CSV.write

tk3369:
This is an orthogonal PR for the same purpose as described in #562.  For background information, please refer to that issue.

### Changes

1. Introduce a new `transform` option for `CSV.write`, which accepts two arguments - column index and value.  This transform function is applied to every cell before writing to the output stream.  The default is `(col, val) -> val`, pretty much an identity function.

2. Throw an error when `nothing` is encountered during `CSV.write`.  The error string looks like this:
```
julia> (c1=[1],c2=[nothing],c3=[2]) |> CSV.write(stdout)
ERROR: A `nothing` value was found in column 2 and it is not a printable value. 
There are several ways to handle this situation:
1) fix the data, perhaps replace `nothing` with `missing`,
2) use `transform` option with a funciton to replace `nothing` with whatever value (including `missing`), or
3) use `Tables.transform` option to transform specific columns
```

### Performance

Performance results indicate that there is **no material degradation** with either the default transform function or a transform function that converts `nothing` to `missing`. The test harness is in `test/perf_write.jl`.

Master:
```
$ git status
HEAD detached at v0.5.22

$ julia --project=. /tmp/perf_write.jl master
Testing without nothing values
10000 rows: Trial(2.778 ms)
100000 rows: Trial(42.258 ms)
1000000 rows: Trial(429.475 ms)
10000000 rows: Trial(5.877 s)
```

This PR:
```
$ julia --project=. /tmp/perf_write.jl new
Testing without nothing values (default transform)
10000 rows: Trial(2.908 ms)
100000 rows: Trial(40.570 ms)
1000000 rows: Trial(415.337 ms)
10000000 rows: Trial(5.623 s)
Testing without nothing values (to-missing transform)
10000 rows: Trial(2.901 ms)
100000 rows: Trial(41.631 ms)
1000000 rows: Trial(421.154 ms)
10000000 rows: Trial(5.829 s)
Testing with nothing values (to-missing transform)
10000 rows: Trial(2.779 ms)
100000 rows: Trial(38.552 ms)
1000000 rows: Trial(380.628 ms)
10000000 rows: Trial(6.017 s)
```

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -186,9 +186,9 @@ function writerow(buf, pos, len, io, sch, row, cols, opts)
         val === nothing && error(
             """
             A `nothing` value was found in column $col and it is not a printable value. 
-            There are several ways to handle this situation:\n
-            1) fix the data, perhaps replace `nothing` with `missing`, \n
-            2) use `transform` option with a funciton to replace `nothing` with whatever value (including `missing`), or\n
+            There are several ways to handle this situation:
+            1) fix the data, perhaps replace `nothing` with `missing`,
+            2) use `transform` option with a funciton to replace `nothing` with whatever value (including `missing`), or
             3) use `Tables.transform` option to transform specific columns
             """)
         posx = writecell(buf, pos[], len, io, val, opts)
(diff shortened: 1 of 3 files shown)


tk3369:
@quinnj have time to review this PR? Thanks

tk3369:
@quinnj I have another use case for this feature. Can we get this merged soon? Thanks

Issue #557, opened 2020-01-05
Specify which columns to use


flip111:
I've read the documentation, but i couldn't find an option to specify which columns to use. Like in pandas [read_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) function there is a `usecols` parameter. Could such a parameter be added?

quinnj:
Implemented in #585

PR #585, opened 2020-03-24
Integration column selection directly with parsing

quinnj:
Implements #557.

We've always had the ability to do fairly efficient column selection
post-parsing on a `CSV.File` object via `Tables.select` (now
`TableOperations.select`), but this still meant parsing the full file,
which, depending on the column type, could be a little costly.

With this PR, we introduce a `select` keyword argument to
`CSV.read`/`CSV.File` which allows passing a list of columns as an
`AbstractVector{Int}`, `AbstractVector{Symbol}`, or
`AbstractVector{String}`, or as a "selector" function of the form `(i,
name) -> keep::Bool` to specify which columns to "keep" after parsing.
Internally, this utilizes a trick we already had in place wherein if a
column type is user-specified as `Missing`, then parsing efficiently
"skips" any type parsing for that column (basically just moves our
parsing state past the cell). So in terms of implementation, this is
surprisingly simple: for any column we don't want to "select", we set
it's column type upfront to `Missing`, parse the file, then remove any
mention of the dropped columns in the final `CSV.File` object.

TODO items left on this PR include:
* Does this API look sound/safe/whole? cc: @nalimilan, @bkamins
* Tests
* Examples

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -198,7 +198,8 @@ Supported keyword arguments include:
   * `use_mmap::Bool=!Sys.iswindows()`: whether the file should be mmapped for reading, which in some cases can be faster
   * `ignoreemptylines::Bool=false`: whether empty rows/lines in a file should be ignored (if `false`, each column will be assigned `missing` for that empty row)
   * `threaded::Bool`: whether parsing should utilize multiple threads; by default threads are used on large enough files, but isn't allowed when `transpose=true` or when `limit` is used; only available in Julia 1.3+
-  * `select`: an `AbstractVector` of `Int`, `Symbol`, or `String`, or a "selector" function of the form `(i, name) -> keep::Bool`; only columns in the collection or for which the selector function returns `true` will be parsed and accessible in the resulting `CSV.File`
+  * `select`: an `AbstractVector` of `Int`, `Symbol`, `String`, or `Bool`, or a "selector" function of the form `(i, name) -> keep::Bool`; only columns in the collection or for which the selector function returns `true` will be parsed and accessible in the resulting `CSV.File`. Invalid values in `select` are ignored.
+  * `drop`: inverse of `select`; an `AbstractVector` of `Int`, `Symbol`, `String`, or `Bool`, or a "drop" function of the form `(i, name) -> drop::Bool`; columns in the collection or for which the drop function returns `true` will ignored in the resulting `CSV.File`. Invalid values in `drop` are ignored.
 * Parsing options:
   * `missingstrings`, `missingstring`: either a `String`, or `Vector{String}` to use as sentinel values that will be parsed as `missing`; by default, only an empty field (two consecutive delimiters) is considered `missing`
   * `delim=','`: a `Char` or `String` that indicates how columns are delimited in a file; if no argument is provided, parsing will try to detect the most consistent delimiter on the first 10 rows of the file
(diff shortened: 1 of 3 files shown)


src/CSV.jl
```diff
@@ -425,6 +428,31 @@ function file(source,
     else
         typecodes = TypeCode[T for _ = 1:ncols]
     end
+    # set any unselected columns to typecode USER | MISSING
+    todrop = Int[]
+    if select !== nothing
+        if select isa AbstractVector{Int}
+            foreach(1:ncols) do i
```
nalimilan: Why not just:
```suggestion
            for i in 1:ncols
```
Same below.

src/CSV.jl
```diff
@@ -425,6 +428,31 @@ function file(source,
...
+                i in select || push!(todrop, i)
+            end
+        elseif select isa AbstractVector{Symbol} || select isa AbstractVector{<:AbstractString}
+            select = map(Symbol, select)
+            foreach(1:ncols) do i
+                ix = findfirst(==(names[i]), select)
+                ix === nothing && push!(todrop, i)
```
nalimilan:
```suggestion
                names[i] in select || push!(todrop, i)
```

src/CSV.jl
```diff
@@ -425,6 +428,31 @@ function file(source,
...
+            end
+        elseif select isa Base.Callable
+            foreach(1:ncols) do i
+                select(i, names[i]) || push!(todrop, i)
+            end
+        else
+            error("`select` keyword argument must be an `AbstractVector` of `Int`, `Symbol`, or `String`, or a selector function of the form `(i, name) -> keep::Bool`")
+        end
+    end
+    sort!(unique!(todrop))
```
nalimilan: This shouldn't be needed since `push!` is only called once for each `i` and in the right order.

bkamins:
Some minor comments (they are not super critical so may be omitted or added later):
1. current implementation is not strict, i.e. it does not error if `select` contains values that are not actually in the file (this is acceptable, but I think it would be good to write in the docs that this is the way it works)
2. Apart from regex, I think that even more useful would be to allow for `Not` so that you can drop columns in this way instead of keeping them
3. Finally normally `AbstractVector{Bool}` is also accepted as a selector in indexing

quinnj:
@nalimilan, yeah, I thought about regexes, but I also thought that's why the "selector" function is so useful because you can do regex, startswith/endswith, and anything else you want.

@bkamins,
1. Yeah, I say we leave it for now, but I'll add a note in the docs; seems like it might be more inconvenient if it was overly strict.
2. Yeah, I went back and forth on this a bit. We could perhaps have a `drop` keyword arg to go along with `select`; that would allow avoiding the extra dependency on InvertedIndices.jl (right? as far as I could tell, CSV.jl would have to take that dependency)
3. Good call; I'll add support for this.

bkamins:
Yes - InvertedIndices.jl would be a dependency then (it is relatively light, but still). As you have said, if you have a function interface it can be worked around. Adding `drop` would be OK (I think what you choose is probably for you to judge - I just say that often it is more convenient to drop columns than select them)

nalimilan:
I'd rather take the dependency on InvertedIndices. If the fact that it's a separate package is an argument not to use `Not`, let's file an issue against Base to include it as that would be a really silly situation. IMO the long-term goal should be that all standard selectors are supported (`Not`, `Between`, `All`, regexes...), possibly thanks to shared helpers in Tables or TableOperations.

quinnj:
I added a `drop` keyword argument; I think I like that vs. `Not` w/ select. I reviewed the DataFrames selection code quite a bit for this, but my feeling is that it's kind of a different use-case; because here we're really just trying to single out a subset of columns, so things like `All` or `Between` don't seem quite as useful/relevant. We could perhaps unify it at some point; but I kind of like the simplicity of this now.

PR #599, opened 2020-04-10
Allow passing types to CSV.Rows

quinnj:
We already allow fairly efficient means of parsing typed values when
iterating via `CSV.Rows` with `Parsers.parse` and `CSV.detect`
definitions, but ideally, you could pass your own types that you care
about to `CSV.Rows` and have them parsed while iterating/consuming the
file.

To do this properly and not cause massive code duplication, it's
required refactoring things quite a bit to share a lot of common logic
between CSV.File and CSV.Rows (hence a previous PR to introduce
CSV.Header). This PR further refactors things so that CSV.File and
CSV.Row use the same core `parserow` method for consuming a single row
of a file.

One potential downside is that we're changing CSV.Rows from using a
single contiguous `Vector{UInt64}` as a tape, to using a
`Vector{Vector{UInt64}`, where each element vector has length 1. It's
not clear to me what kind of memory cache line penalty we might pay, but
I plan on doing some benchmarking to try and see if it's drastic
(computers are pretty smart these days, so hopefully it's fine).

Otherwise, the changes here aren't too drastic; mainly the refactoring
and then the "value access" functions changing from always returning
Strings, to returning the typed value if there was one.

I need to add some additional tests here, and generally benchmark/kick
the tires to see if anything else shakes out, but seems to work fairly
well so far.

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -439,73 +439,13 @@ end # @static if VERSION >= v"1.3-DEV"
     return perthreadrows, perthreadtapes, refs, typecodes, intsentinels
 end
 
-function parsetape(::Val{transpose}, ncols, typemap, tapes, poslens, buf, pos, len, limit, positions, pool, refs, lastrefs, rowsguess, typecodes, intsentinels, debug, options::Parsers.Options{ignorerepeated}) where {transpose, ignorerepeated}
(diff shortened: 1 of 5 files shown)


quinnj:
Ok, investigated performance a bit and it looks like there's a 10% slowdown for `CSV.Rows`, which seems acceptable to me for now, given the increase in functionality.

Issue #608, opened 2020-04-17
Support generating lines of CSV as an iterator


iamed2:
It came up in https://github.com/invenia/LibPQ.jl/pull/172 that it would be useful to have a version of `CSV.write` that generates rows iteratively. So this new `CSV.iterwrite` function would work like:

```julia
collect(CSV.iterwrite(table)) == readlines(seekstart(CSV.write(IOBuffer(), table)))
```

This seems pretty difficult to shoehorn in to the performance-optimized `CSV.write` code though.

quinnj:
Initial PR to support this: https://github.com/JuliaData/CSV.jl/pull/611

caseykneale:
Oh wow you beat me too it Eric - I have a similar but related issue. I ran into this problem at work in a bad way.

I had a 100k CSV with ~2k columns. The file could not be saved with CSV.write - it crashed Julia over and over despite having plenty of memory.. 

I was able to use eachrow() and write using delimitedfiles.jl in <1 min

Not the biggest invoncenience, but there are edge cases which freeze CSV.jl completely and requires me to reload a large dataset and try again.

quinnj:
This is merged into master

PR #611, opened 2020-04-18
Provide a row writing iterator

quinnj:
Implements suggestion in #608.

This introduces a new structure `CSV.RowWriter`, which is the writing
complement to `CSV.Rows`. It allows taking any Tables.jl-compatible
input, with the same options as `CSV.write`, but instead of writing out
to a file, returns an iterator, where elements are a full csv-written
row.

We mostly had all the machinery in place to provide this, so it's not a
ton of new code. I wondered a bit if the first iteration should produce
column names or just data and if that should be configurable, but
decided to just do column names on first iteration and data thereafter.
I think it's easy enough for callers to drop the first iteration if
not needed anyway, but I'd love to hear people's feedback on that.

I haven't done any tests or docs for this yet; mainly put it up in order
for people to review, so let me know what you think of the design and if
we're missing anything major here.

cc: @iamed2

### DIFF
--- a/src/write.jl
+++ b/src/write.jl
@@ -46,7 +46,13 @@ struct RowWriter{T, S, O}
     schema::S
     options::O
     buf::Vector{UInt8}
+    header::Vector
 end
+
+Base.IteratorSize(::Type{RowWriter{T, S, O}}) where {T, S, O} = Base.IteratorSize(T)
+Base.length(r::RowWriter) = length(r.source) + 1
+Base.size(r::RowWriter) = (length(r.source) + 1,)
+Base.eltype(r::RowWriter) = String
 
 function RowWriter(table;
     delim::Union{Char, String}=',',
@@ -61,6 +67,7 @@ function RowWriter(table;
     missingstring::AbstractString="",
     transform::Function=(col,val) -> val,
     bom::Bool=false,
+    header::Vector=String[],
     bufsize::Int=2^22)
     checkvaliddelim(delim)
     (isascii(something(openquotechar, quotechar)) && isascii(something(closequotechar, quotechar)) && isascii(escapechar)) || throw(ArgumentError("quote and escape characters must be ASCII characters "))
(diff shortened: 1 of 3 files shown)


iamed2:
A `writeheader` option would be useful; could `iterate(r::RowWriter)` call `iterate(r, (state, cols))` if `!writeheader`?

src/write.jl
```diff
@@ -41,6 +42,90 @@ tup(x::AbstractString) = Tuple(codeunits(x))
...
+    oq, cq = openquotechar !== nothing ? (openquotechar % UInt8, closequotechar % UInt8) : (quotechar % UInt8, quotechar % UInt8)
+    e = escapechar % UInt8
+    opts = Options(tup(delim), oq, cq, e, tup(newline), decimal % UInt8, dateformat, quotestrings, tup(missingstring), transform, bom)
+    source = Tables.rows(table)
+    sch = Tables.schema(source)
+    return RowWriter(source, sch, opts, Vector{UInt8}(undef, bufsize), header)
+end
+
+struct DummyIO <: IO end
+Base.write(io::DummyIO, a::SubArray{T,N,<:Array}) where {T,N} = error("`bufsize` for `CSV.RowWriter` was too small (default 4MB); try again passing a larger value for `bufsize`")
```
iamed2, line 104: This is a very smart quick fix, I like it

iamed2:
If `newline='\n'` we get strings that end with `'\n'` and if `newline=""` we get strings which have no newline ending, is that correct? That's my ideal.

quinnj:
> A `writeheader` option would be useful; could `iterate(r::RowWriter)` call `iterate(r, (state, cols))` if `!writeheader`?

That's a clever idea; I'll try it out.

> If `newline='\n'` we get strings that end with `'\n'` and if `newline=""` we get strings which have no newline ending, is that correct? That's my ideal.

Yes, that's the expected behavior; I've made the possible and added a test.

PR #615, opened 2020-04-19
Add CSV.chunks for getting individual thread chunks of CSV.File{true}

quinnj:


xiaodaigh:
Just tried it. It's amazing. Are the chunks *guaranteed* to have the same column types or it can change?

quinnj:
> Just tried it. It's amazing. Are the chunks *guaranteed* to have the same column types or the type for each chunk may differ?

Guaranteed to be the same; we run the “thread merging” algorithm after parsing to ensure everything is the same across chunks.

xiaodaigh:
After testing it abit, I think that it loads the whole dataset before chunking. Hmm, I wonder if it's possible to read only a chunk of the file, infer type, and return a dataframe and then move to the next chunk. This way saves RAM but the types may differ between chunks as earlier chunks are inferred using less data.

xiaodaigh:
I think something building on the below is what I was looking for. Performance is not bad.

```julia
read_chunks(name, size = 2^20) = begin
    res = []
    file = open(name, "r")
    upto = 0
    header = Symbol[]
    while true
        aa = read(file, size)
        faa = findlast(x->x==newline, aa)

        if isnothing(faa)
            return res
        end
        if upto == 0
            df = CSV.read(IOBuffer(@view aa[1:faa]))
            header = names(df)
        else
            df = CSV.read(IOBuffer(@view aa[1:faa]), header=header)
        end
        upto += faa

        seek(file, upto)
        position(file)
        push!(res, df)
    end

    res
end
```

quinnj:
I have an idea for a better approach here.

PR #602, opened 2020-04-15
Port to CategoricalArrays 0.8

nalimilan:
The change to `levels!` implies that all reference codes will be updated at the end to match the new order of levels (unless they were already sorted).

Requires the yet-to-be released CategoricalArrays 0.8.

PR #617, opened 2020-04-20
Update Project.toml

bkamins:
Changes:

1. Update package version to 0.6.2 (it was not bumped to 0.6.1 when last release was made). I moved to 0.6.2 as there are some changes on master since 0.6.1 release
2, Allow CategoricalArrays 0.8 (this is a crucial point why this release is required now)
3. Allow DataFrames 0.21 (it is not released yet, but will be released very soon, so I think it is good to allow it already to avoid another release in a few days)

CC @nalimilan

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CSV"
 uuid = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 authors = ["Jacob Quinn <quinn.jacobd@gmail.com>"]
-version = "0.6.0"
+version = "0.6.2"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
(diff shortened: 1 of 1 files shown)


bkamins:
@quinnj - also maybe it is best if you decide how to update https://github.com/JuliaRegistries/General/blob/master/C/CSV/Compat.toml.

Currently due to loose dependency version restrictions for old versions of CSV.jl we have a typical situation where a very old version of CSV.jl gets installed in case of version conflict (e.g. with CategoricalArrays.jl v0.8 a version 0.3.1. of CSV.jl gets selected)

nalimilan:
Just stating the obvious: https://github.com/JuliaData/CSV.jl/pull/602/files needs to be merged first.

bkamins:
Ah - sorry, I have not noticed. Then of course this is obsolete, so closing this one.

Issue #522, opened 2019-10-27
Allow user-provided row and field error callback functions for custom error handling


quinnj:

quinnj:
@NHDaly, @mdashti, I'm circling back round to what we want here: what I'm thinking is allowing the user to provide callbacks like `invalidrow=rownumber -> ...`, which would be executed [here](https://github.com/JuliaData/CSV.jl/blob/1eddc33c1f063785a4d3d6c48c7e90efd1d058ad/src/file.jl#L533) and [here](https://github.com/JuliaData/CSV.jl/blob/1eddc33c1f063785a4d3d6c48c7e90efd1d058ad/src/file.jl#L552) (if a row has too few or too many columns), and `invalidcell=(row, col, T) -> ...` that would be executed [here](https://github.com/JuliaData/CSV.jl/blob/1eddc33c1f063785a4d3d6c48c7e90efd1d058ad/src/file.jl#L737) and [here](https://github.com/JuliaData/CSV.jl/blob/1eddc33c1f063785a4d3d6c48c7e90efd1d058ad/src/file.jl#L788). 

Things I'm wondering about:
* are there any other arguments we should provide to the callback functions? Just row number for `invalidrow` and row/column number and type for `invalidcell`; would anything else be useful? Maybe some kind of error code? We have the underlying error codes from Parsers.jl that we could maybe surface
* Currently for invalid rows, we fill in invalid cells with `missing`; I'm thinking we would keep this behavior, so the user-provided callback would just be able to do their own side-effects elsewhere
* For invalid cells, we currently have the behavior that if the user provided the column's type, then we force-promote the column type to `Union{Missing, T}` and fill in the invalid cell with `missing`; if the column type was inferred, then we promote the column type to String and move along. I'm trying to think through how exactly we integrate `invalidcell` callback, here are questions I have:
  * would we execute the callback if user didnt' provide column type and we're just promoting to String? I think not since you've kind of opted for CSV.jl to figure out the types, so a "type error" is more of an internal promotion thing rather than something the user will care about
  * should we care if a user provided the `strict` or `silencewarnings` keyword arguments? If they passed `strict=true`, then I think we still need to throw an error, which seems like we don't need to invoke the callback as well. So my thinking is that we'd only invoke the `invalidcell` callback if the user provided a column type explicitly, didn't pass `strict=true`, and we encountered an error parsing the provided type. We would still fill that cell in with `missing` value though.

[...]

PR #618, opened 2020-04-21
Allow custom callbacks for invalid row/cell

quinnj:
This implements #522. It allows providing two different callbacks:

* `invalidrow`: a callback that takes a single `Integer` representing
the row number where an invalid state was detected (not enough columns
parsed or more columns than expected were detected)

* `invalidcell`: a callback called when an error occurs parsing a
user-provided type (Int, Float64, Bool, Date, DateTime, or Time); the
callback should take 5 arguments: 1) the type attempted while parsing,
2) a string of the cell where parsing failed, 3) a Parsers.jl `code`
return code value that encodes events/states that occurred while
parsing, 4) the row the invalid cell was on, and 5) the column the
invalid cell was on.

Providing an `invalidrow` or `invalidcell` function forces
`silencewarnings=true` for invalid rows or cells, respectively, i.e. no
warning will be printed if an error callback function is provided.

If `strict=true` is provided, `invalidcell` will not be invoked.

This PR includes a simple test, but no docs yet.

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -204,6 +204,8 @@ function File(source;
     pool::Union{Bool, Real}=0.1,
     strict::Bool=false,
     silencewarnings::Bool=false,
+    invalidrow=nothing,
+    invalidcell=nothing,
     debug::Bool=false,
     parsingdebug::Bool=false,)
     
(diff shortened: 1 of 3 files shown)


quinnj:
Closing in favor of the approach in #656

PR #619, opened 2020-04-21
Reading CSV with boolean column

### LINKED ISSUE #596: Reading CSV with boolean column
hi I was reading data which contains boolean values. Here is an example:

no | name | active | rain | water | winddir | windlevel | temp
-- | -- | -- | -- | -- | -- | -- | --
502 | Gdynia ul. Inżynierska | True | True | False | False | False | False
503 | Gdynia ul. Janki Bryla | True | True | False | False | False | False

The problem is that the columns like `active` `rain` etc. are not recognised as having boolean values. They have string type.
When I read the same file using Python Pandas it looks ok:
```
 #   Column     Non-Null Count  Dtype 
---  ------     --------------  ----- 
 0   no         107 non-null    int64 
 1   name       107 non-null    object
 2   active     107 non-null    bool  
 3   rain       107 non-null    bool  
 4   water      107 non-null    bool  
 5   winddir    107 non-null    bool  
 6   windlevel  107 non-null    bool  
 7   temp       107 non-null    bool  
 8   pressure   107 non-null    bool  
 9   humidity   107 non-null    bool  
 10  sun        107 non-null    bool  
dtypes: bool(9), int64(1), object(1)
memory usage: 2.7+ KB
```

So the problem is not with the file.
Do I miss something?

klangner:
Added alternative bool values as default values when read CSV from file.

Fixes #596


quinnj:
Thanks for this @klangner !

src/file.jl
```diff
@@ -194,8 +194,8 @@ function File(source;
     dateformat::Union{String, Dates.DateFormat, Nothing}=nothing,
     dateformats::Union{AbstractDict, Nothing}=nothing,
     decimal::Union{UInt8, Char}=UInt8('.'),
-    truestrings::Union{Vector{String}, Nothing}=nothing,
-    falsestrings::Union{Vector{String}, Nothing}=nothing,
+    truestrings::Union{Vector{String}, Nothing}=["true", "True", "TRUE"],
```
nalimilan, line 197: What's the point of still allowing `nothing` if it's neither the default nor documented?

klangner: I'm not sure if this is undefended, but you can still provide nothing and get default behaviour . But probably changing the type here is better. Should I do it?

nalimilan: The effect of passing `nothing` is actually determined by Parsers.jl:
https://github.com/JuliaData/Parsers.jl/blob/545326062542da983a2121236f176b0f5ccc5714/src/Parsers.jl#L18-L19

I'd say we should either disallow passing `nothing` here, or change Parsers.jl to use the same default as here. AFAICT, the point of passing `nothing` is that it uses [an optimized path](https://github.com/JuliaData/Parsers.jl/blob/c7199940a0d412ae9b72e8f933fa5909fe56ddf2/src/bools.jl#L8) for performance: the default values for CSV.jl should definitely be optimized for, so maybe changing Parsers.jl is the best option @quinnj?

PR #632, opened 2020-05-27
Bump [compat] FilePathsBase="0.6,0.7,0.8,0.9"

ablaom:
We need this to eliminate package conflicts at test time in MLJBase and related packages.

### DIFF
--- a/test/write.jl
+++ b/test/write.jl
@@ -1,4 +1,6 @@
 using CSV, Dates, WeakRefStrings, CategoricalArrays, Tables
+using FilePathsBase
+using FilePathsBase: /
 
 const default_table = (col1=[1,2,3], col2=[4,5,6], col3=[7,8,9])
 
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ WeakRefStrings = "ea10d353-3f73-51f8-a26c-33c1cb351aa5"
 [compat]
 CategoricalArrays = "0.8"
 DataFrames = "0.21"
-FilePathsBase = "0.6,0.7,0.8"
+FilePathsBase = "0.6,0.7,0.8,0.9"
 Parsers = "1"
 PooledArrays = "0.5"
 Tables = "1"


test/write.jl
```diff
@@ -1,4 +1,6 @@
 using CSV, Dates, WeakRefStrings, CategoricalArrays, Tables
+using FilePathsBase
```
quinnj, line 2: Why is this needed?

ablaom: Possibly you only need the line after: `using FilePathsBase: /`. Without that the FilePaths test fail.

quinnj: @oxinabox @rofinn , do either of you know why this would now be needed? It seems odd to me. Did FilePathsBase stop exporting something or something?

rofinn: Yes, `/` and `join` no longer extend the base methods as requested by some core dev team members. If you want those methods to work with a path type then you need to do `using FilePathsBase: /, join`. FWIW, we might be able to think of a way for FilePathsBase to be a test-only dependency if we loosen a couple of type restrictions?

ExpandingMan:
Didn't notice this PR when I wrote #636.  This issue is really wreaking havoc for me.

I notice this is lacking tests for 1.4.  Could we add those and then merge this?

rofinn:
I'm fine with merging this as is if Jacob Quinn doesn't have any objections in the next day or so. I think we should probably try to drop the primary FilePathsBase dependency and loosen the type restrictions, but that can happen later.

oxinabox:
We had that before with no dependency and it working.
Then FilePathBase changed something and it broke.
And the package manager didn't know since there was no dependency.

Then when the fix was made a dependency was added.

rofinn:
I’m guessing that was when AbstractPath stopped subtyping AbstractString, as this package now just converts to a string. There’s probably a better way to handle that and we should still continue testing against FilePathsBase.

oxinabox:
> I’m guessing that was when AbstractPath stopped subtyping AbstractString, a

No it was after that because 
https://github.com/JuliaData/CSV.jl/pull/405 was made when FilePaths stopped subtyping String (0.5)

where as the PR that added the dependency was:
https://github.com/JuliaData/CSV.jl/pull/511/

ablaom:
FYI The only CI fail is a time-out on Windows julia 1.0

```
curl: (7) Failed to connect to storage.googleapis.com port 443: Timed out
```

Issue #431, opened 2019-05-10
Provided types only partially interpreted?


Drvi:
Hi and thanks for the awesome package!

I was just upgrading to a new version of CSV and I noticed that the provided type info is not being interpreted exactly:
```julia
julia> DataFrame(CSV.File(IOBuffer("a,b\n1,2"), types = [Float32,UInt8]))
1×2 DataFrame
│ Row │ a       │ b     │
│     │ Float64 │ Int64 │
├─────┼─────────┼───────┤
│ 1   │ 1.0     │ 2     │
```

I do a lot of these explicit type conversions to save some space when working with large tables.

Please let me know if I can help with this..

CSV v0.5.2
Julia 1.1.0

quinnj:
Yes, I was anticipating this would come up. The issue here is that during parsing we treat everything as Int64/Float64, and any input types are just converted to those. I think we could have a way to pass an `inttype=Int8` and/or `floattype=Float32` arguments that would control how int/float columns are handled. But maybe we could still figure out how to propagate the types to the resulting columns as well.

nalimilan:
Disregarding how the data is stored internally, the `types` argument should be respected IMO.

versipellis:
This is coming up as an issue for me.

I have a CSV with columns that require Int128, and this error gets thrown:
```
warnings: error parsing Int64 on row = 2, col = 1: "10000000000000000000000000000000,", error=INVALID: DELIMITED | INVALID_DELIMITER | OVERFLOW 
```

quinnj:
Yeah, the case of `Int128` or `BigInt` are not ideal right now. You'd be better off doing something like:
```julia
df = CSV.File(filename; types=Dict(bigintcol=>String)) |> Tables.transform(bigintcol=x->Parsers.parse(BigInt, x)) |> DataFrame
```

ExpandingMan:
Note that currently if you write e.g. `UInt64`, these will look like integers in the file, and by default it tries to read them back as `Float64`.  If you specify `types=Dict(col=>UInt64)` it apparently tries to read as `Int` and can get an overflow.

sbromberger:
It'd be really great to get this working. For performance reasons, `UInt32` is an ideal data type for 99% of my work. I'd love to be able to read in UInt32s directly.

quinnj:
Fixed in https://github.com/JuliaData/CSV.jl/pull/649

Drvi:
Amazing! Thank you.

PR #639, opened 2020-06-10
Refactor CSV internals to produce fully mutable columns by default

quinnj:
This involves quite a bit, so I'll try to spell out some of the chunks
of work going on here, starting with the key insight that kicked off all
this work.

Previously, to get a "type-stable" inner parsing loop over columns, I
had come to the conclusion you actually needed the same concrete typed
column type, no matter the column type of the data. This led to the
development of the `typecode` + `tapes` system that has served us pretty
well over the last little while. The idea there was any typed value
could be coerced to a `UInt64` and thus every column's underlying
storage was `Vector{UInt64}`, wrapped in our `CSV.Column` type to
reinterpret the bits as `Int64`, `Float64`, etc. when indexed. This was
a huge step forward for the package because we were no longer trying to
compile custom parsing kernels for every single file with a unique set
of column types (which could be fast once compiled, but with considerable
overhead at every first run).

The disadvantage of the tape system was it left our columns read-only;
yes, we could have made the regular bits types mutable without too much
trouble, but it was hard to generalize to all types and we'd be stuck
playing the "make `CSV.Column` be more like `Array` in _this_ way" game
forever. All those poor users to have tried mutating operations on
`CSV.Column` not realizing they needed to make a copy.

While refactoring ODBC.jl recently, at one point I realized that with
the fixed, small set of unique types you can receive from a database,
it'd be nice to roll my own "union splitting" optimization and unroll
the 10-15 types myself in order to specialize. This is because the core
Julia provided union-splitting algorithm will bail once you hit 4 unique
types. But in the case of ODBC.jl, I knew the static set of types and
unrolling the 12 or so types for basically `setindex!` operations could
be a boon to performance. As I sat pondering how I could reach into the
compiler with generated functions or macros or some other nonsense, the
ever heroic Tim Holy swooped in on my over-complicated thinking and
showed me that just writing the `if` branches and checking the objects
against concrete types _is exactly the same_ as what union-splitting
does. What luck!

What this means for CSV.jl is we can now allocate "normal" vectors and
in the innermost parsing loop (see `parserow` in `file.jl`), pull the
column out of our `columns::Vector{AbstractVector}` and check it against
the standard set of types we support and manually dispatch to the right
parsing function. All without spurious allocations or any kind of
dynamic dispatch. Brilliant!

In implementing this "trick", I realized several opportunities to
simplify/clean things up (net -400 LOC!), which include:
* Getting rid of `iteration.jl` and `tables.jl`; we only need a couple
lines from each now that we are getting rid of `CSV.Column` and the
threaded-row iteration shenanigans we were up to
* Getting rid of the `typecodes` we were using as our pseudo-type
system. While this technically could have stayed, I opted to switch to
using regular Julia types to streamline the process from parsing ->
output, and hopefully pave the way to supporting a broader range of
types while parsing (see #431).
* Move all the "sentinel array" tricks we were using into a more
formally established (and tested)
[SentinelArrays.jl](https://github.com/JuliaData/SentinelArrays.jl)
package; a `SentinelArray` uses a sentinel value of the underlying array
type for `missing`, so it can be thought of as a `Vector{Union{T,
Missing}}` for all intents and purposes. This package's array types will
be used as solutions to using regular `Vector{Union{T, Missing}}` are
worked out; specifically, how to efficiently allow a non-copying
operation like `convert(Vector{T}, A::Vector{Union{T, Missing}})`

Probably most surprisingly with this PR is that it's pretty much
non-breaking! I haven't finished the `CategoricalArrays` support just
yet, but we'll include it for now as we get ready to make other
deprecations in preparation for 1.0. The other bit I'm still mulling
over is how exactly to treat string columns; in this PR, we just fully
materialize them as `Vector{String}`, but that can be a tad expensive
depending on the data. I'm going to try out some solutions locally to
see if we can utilize `WeakRefStringArray` but will probably leave the
option to just materialize the full string columns, since we've had
people ask for that before.

If you've made it this far, congratulations! I hope it was useful in
explaining a bit about what's going on here. I'd love any
feedback/thoughts if you have them. I know it's a lot to ask anyone to
review such a monster PR in their free time, but if you're willing, I'm
more than happy to answer questions or chat on slack (#data channel) to
help clarify things.


src/utils.jl
```diff
@@ -1,10 +1,10 @@
-# PooledString is an internal-only type for efficiently tracking string data + length
-# all strings indexed from a column/row will always be a full String
-# specifically, it allows avoiding materializing full Strings for pooled string columns while parsing
-# and allows a fastpath for materializing a full String when no escaping is needed
 export PooledString
 struct PooledString <: AbstractString end
 
+# PointerString is an internal-only type for efficiently tracking string data + length
+# all strings indexed from a column/row will always be a full String
+# specifically, it allows avoiding materializing full Strings for pooled string columns while parsing
```
nalimilan, line 6: Is it the case that `PointerString` now indicates that strings should be pooled? Then what is `PooledString` for?

PR #656, opened 2020-06-25
Start work on supporting filtering while parsing

quinnj:


quinnj:
@nhdaly, sorry for the slightly random ping, but I'm actually diving back in to the code here and want to make a push. would you mind taking a look here if you have a moment? Happy to jump on a quick call too if that'd be easier to do a quick review of what's going on here.

PR #662, opened 2020-06-26
Deprecate the categorical keyword argument

quinnj:


src/file.jl
```diff
@@ -191,7 +191,7 @@ function File(source;
     type=nothing,
     types=nothing,
     typemap::Dict=Dict{Type, Type}(),
-    categorical::Union{Bool, Real}=false,
+    categorical=nothing,
```
nalimilan, line 194:
```suggestion
    categorical::Union{Bool, Real, Nothing}=nothing,
```

src/rows.jl
```diff
@@ -117,7 +117,7 @@ function Rows(source;
     type=nothing,
     types=nothing,
     typemap::Dict=Dict{Type, Type}(),
-    categorical::Union{Bool, Real}=false,
+    categorical=nothing,
```
nalimilan, line 120:
```suggestion
    categorical::Union{Bool, Real, Nothing}=nothing,
```

src/header.jl
```diff
@@ -94,6 +94,9 @@ getdf(x::AbstractDict{Int}, nm, i) = haskey(x, i) ? x[i] : nothing
     (types !== nothing && any(x->!isconcretetype(x) && !(x isa Union), types isa AbstractDict ? values(types) : types)) && throw(ArgumentError("Non-concrete types passed in `types` keyword argument, please provide concrete types for columns: $types"))
     checkvaliddelim(delim)
     ignorerepeated && delim === nothing && throw(ArgumentError("auto-delimiter detection not supported when `ignorerepeated=true`; please provide delimiter like `delim=','`"))
+    if categorical !== nothing
+        @warn "the `categorical` keyword argument is deprecated; in the next release, you'll need to do `using CategoricalArrays; catg = categorical(f.pooled_column)`, where `pooled_column` is the column of a `CSV.File` you want as a `CategoricalArray`"
```
nalimilan, line 98: Would something like `typemap=Dict(String => CategoricalValue{String}})` work too?

quinnj: No; there won't be any way to special case `CategoricalValue` inside CSV.jl anymore. With that proposal, it would try to call `Base.parse(CategoricalValue{String}, str)` for each string cell.

nalimilan: Why would a special case be needed? Aren't columns allocated using `Tables.allocatecolumn`?

quinnj: No, we don't use `Tables.allocatecolumn` because we require specialized array types (`SentinelArray`, `ChainedVector`). Also, we specialize "pooled" columns, reading them as a `Vector{UInt32}` and the `categorical` keyword was just changing how we "wrapped up" that vector after parsing finished.

PR #665, opened 2020-06-27
Add CSV.Chunks for iterating over chunks of large files

quinnj:


quinnj:
Cc: @xiaodaigh , fyi, since we've talked about chunking before

xiaodaigh:
nice. But it's crashing on julia v1.4.2 when trying to read the largest Fannie Mae performance data (13G in size). I have 64RAM and it seems to eat up all the RAM

quinnj:
@xiaodaigh, I don't see any error on the latest release (0.7.2) on the same dataset; my machine only has 32gb ram. Could you try again and see if you get a crash again?

PR #681, opened 2020-07-08
fix for issue#679 wrt select/drop and types options

### LINKED ISSUE #679: error with combination of types and select/drop options of CSV.File
An error occurs whenever the `types` option to `CSV.File` contains a pair corresponding to a dropped column with a type `<: Union{Int16, Int32, Int8, UInt16, UInt32, UInt8}` (aka `SmallIntegers` within `CSV.jl`).

MWE:
file `test.csv`:
```
"a","b","c"
1,"a",12
1,"a",12
```

```julia
julia> CSV.File("test.csv"),select=["a"],types=Dict("a"=>Int32,"b"=>String,"c"=>Int16))
ERROR: MethodError: Cannot `convert` an object of type Missing to an object of type Int16
Closest candidates are:
  convert(::Type{T}, ::T) where T<:Number at number.jl:6
  convert(::Type{T}, ::Number) where T<:Number at number.jl:7
  convert(::Type{T}, ::Ptr) where T<:Integer at pointer.jl:23
  ...
Stacktrace:
 [1] setindex!(::Array{Int16,1}, ::Missing, ::Int64) at ./array.jl:826
 [2] copyto!(::Array{Int16,1}, ::Array{Union{Missing, Int16},1}) at ./multidimensional.jl:962
 [3] Array at ./array.jl:541 [inlined]
 [4] convert(::Type{Array{Int16,1}}, ::Array{Union{Missing, Int16},1}) at ./array.jl:533
 [5] CSV.File(::CSV.Header{false,Parsers.Options{false,false,true,false,Missing,UInt8,Nothing},Array{UInt8,1}}; startingbyteposition::Nothing, endingbyteposition::Nothing, limit::Int64, threaded::Nothing, typemap::Dict{Type,Type}, tasks::Int64, debug::Bool) at /home/kragol/.julia/dev/CSV/src/file.jl:280
 [6] CSV.File(::String; header::Int64, normalizenames::Bool, datarow::Int64, skipto::Nothing, footerskip::Int64, transpose::Bool, comment::Nothing, use_mmap::Nothing, ignoreemptylines::Bool, select::Array{String,1}, drop::Nothing, missingstrings::Array{String,1}, missingstring::String, delim::Nothing, ignorerepeated::Bool, quotechar::Char, openquotechar::Nothing, closequotechar::Nothing, escapechar::Char, dateformat::Nothing, dateformats::Nothing, decimal::UInt8, truestrings::Array{String,1}, falsestrings::Array{String,1}, type::Nothing, types::Dict{String,DataType}, categorical::Nothing, pool::Float64, lazystrings::Bool, strict::Bool, silencewarnings::Bool, debug::Bool, parsingdebug::Bool, kw::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/kragol/.julia/dev/CSV/src/file.jl:215
 [7] top-level scope at REPL[37]:1

```
Note: the error may look a bit different when multi-threading is enabled but is essentially the same.

A simple workaround is to remove the pairs in the `types` `Dict` that are supposed to be dropped according to the `select` or `drop` options.

```
julia 1.4.2
CSV 0.7.2
```

kragol:
Fixes #679 
This may not be the cleanest fix but it seems to work. I guess it can be fine for now since the whole parsing is apparently in the process of being rewritten.

Note: I am a Julia beginner so any feedback is appreciated.

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -273,7 +273,7 @@ function File(h::Header;
                 end
             else
                 # if no missing values were parsed for a column, we want to "unwrap" it to a plain Vector{T}
-                if !anymissing(flags[i])
+                if !anymissing(flags[i]) & (i ∉ h.todrop)
                     if column isa Vector{Union{Missing, Bool}}
                         columns[i] = convert(Vector{Bool}, column)
                     elseif types[i] !== Union{} && types[i] <: SmallIntegers


quinnj:
Thanks for reporting the issue @kragol! And proposing a fix! While this fix is fine, there's actually a slightly simpler fix that I've put up in https://github.com/JuliaData/CSV.jl/pull/683. We have a `flags` array that carries various bit flags that get/set properties for columns, one of which is if `anymissing` values were parsed. So my alternative fix is to just make sure when a column is marked as `WILLDROP`, we also mark `ANYMISSING` since the entire column will be treated as `missing` values.

PR #684, opened 2020-07-09
Fix typemap for our new custom types world

### LINKED ISSUE #680: weird typemap behavior (bug?)
The `typemap` option to `CSV.File` produces incorrect types when mapping Integer types (I haven't really tested other types though). Basically, all `Int` columns are converted to `Float64` whatever the typemap indicates as a target type.

MWE:
file `test.csv`:
```
"a","b","c"
1,"a",12
1,"a",12
```

```julia
julia> CSV.File("test.csv"),typemap=Dict(Int64=>Int32))
2-element CSV.File{false}:
 CSV.Row: (a = 1.0, b = "a", c = 12.0)
 CSV.Row: (a = 1.0, b = "a", c = 12.0)
```
Here, the values of `a` and `c` are converted to `Float64` instead of `Int32` as requested.

```
julia 1.4.2
CSV 0.7.2
```

quinnj:
Fixes #680. Before custom types, the `typemap` keyword argument was
really only about mapping between the standard, supported types. With
custom types, we still only support certain type mappings (Int to Float,
Any type to String), but we also want to support type mappings like
`Int64 => Int32`. This PR readjusts how typemap works when detecting
column types to account for the possiblity of custom Integer or
AbstractFloat type mappints for Int64 & Float64, and moving directly to
String if that's specified.

### DIFF
--- a/src/chunks.jl
+++ b/src/chunks.jl
@@ -70,7 +70,7 @@ function Chunks(source;
     parsingdebug::Bool=false)
 
     limit=typemax(Int64)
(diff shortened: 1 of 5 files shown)


kragol:
Wow that was fast. Thanks a lot for the quick fix and the detailed explanation.

PR #685, opened 2020-07-09
Add docs for reading non-UTF-8 files

nalimilan:
It could make sense for StringEncodings to allow `read(f, enc"ISO-8859-1")` as a shorter variant of `open(read, f, enc"ISO-8859-1")` (https://github.com/JuliaStrings/StringEncodings.jl/pull/37). Currently all functions that decode return strings rather than vectors of bytes, but that would probably be OK. Anyway, for CSV.jl it's better to document something that works with older releases.


kragol:
You might also want to add a caveat about the performance. 

I recently tried to read some 4GB .csv non UTF-8 files using `StringEncodings`: it was roughly 5X to 10X slower than reading the same file after it had been statically converted to UTF-8. Best part was that statically converting the file to UTF-8 (using `iconv` on Linux) and then reading it with `CSV.File` was still way faster than reading it as non UTF-8 by means of `StringEncodings`. 

Of course I guess one cannot always statically convert to UTF-8 but it clearly is the better option at the moment when it is possible.

quinnj:
This is great; thanks @nalimilan. @kragol, I think the difference in timings your seeing is just the actual conversion cost; when you do `CSV.File(open(file, enc"whatever"))` it has to first read the entire file in `whatever` encoding, then do the csv parsing. So to do a more fair comparison, you'd wand to time doing the `iconv` re-encoding, then reading `CSV.File`.

nalimilan:
Actually I had not done any performance optimization in StringEncodings, and it was really slow. Profiling showed that most of the time was in calling `readbytes!` on the underlying data, due to a lack of a specialized method. With https://github.com/JuliaStrings/StringEncodings.jl/pull/38 it's about 10 times faster, and in my test it's "only" about twice slower than parsing a UTF-8 CSV file.

quinnj:
Ah, that's great to hear!

PR #686, opened 2020-07-10
Fix case when type is provided for dropped columns

quinnj:
Additional fix for #679. There was an alternative issue discovered in
 #679 where a type was given for a dropped column, like `CSV.File(file;
 types=Dict(:id=>Int32), select=["first"])`. During parsing, it knew we
 would drop the `id` column, so it parsed it as `Missing`, but then
 tried to "set" the value to a `Vector{Int32}` which threw an error.
 When dropping a column, we can basically ignore any provided type
 information instead, so in the `Header` routine, we make sure to mark
 the type of a column as `Missing` if it will be dropped.

### DIFF
--- a/src/header.jl
+++ b/src/header.jl
@@ -279,6 +279,7 @@ getdf(x::AbstractDict{Int}, nm, i) = haskey(x, i) ? x[i] : nothing
     end
     for i in todrop
         flags[i] |= WILLDROP | ANYMISSING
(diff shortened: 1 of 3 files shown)


kragol:
Thanks for this fix. As far as I can tell, it also fixes the issue in single-threaded mode, which makes your previous patch #683 unnecessary.

PR #687, opened 2020-07-10
Change deprecation warning for CSV.read

quinnj:
Fixes https://github.com/JuliaData/DataFrames.jl/issues/2309. After much
discussion, people really like the `CSV.read` function call naming, so
it was decided to keep it around, while still allowing a break from
DataFrames.jl as a dependency.

I also realized that `CSV.read` does indeed provide one bit of
value/functionality: we can wrap `CSV.File` in `Tables.CopiedColumns`.
What this means is that columnar sinks can safely use the columns passed
without needing to make copies; i.e. they can assume ownership of the
columns. In `CSV.read`, the user is essentially saying, "I want to make
a `CSV.File` and pass it directly to `sink`" which also implies that
`CSV.File` doesn't need to "own" its own columns.

The only question left in my mind is, with the 1.0 release, what to do
with `CSV.read(file)` when no `sink` argument is passed. Suggestions
have included just returning a `CSV.File`, since it's a valid table
anyway. Or allowing DataFrames.jl to define the no-sink-arg version and
return a `DataFrame`; that one's a bit awkward as a form of "blessed"
type piracy, but could also be useful for users as convenience (they
just have to remember to do `using DataFrames` before trying to use it).
The other awkward part is that we're currently warning users of the
deprecation and that they should explicitly spell out `DataFrame`
whereas we might not actually require that.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -30,18 +30,19 @@ include("rows.jl")
 include("write.jl")
 
 """
-`CSV.read(source; copycols::Bool=false, kwargs...)` => `DataFrame`
+`CSV.read(source, sink::T; kwargs...)` => T
 
-Parses a delimited file into a `DataFrame`. `copycols` determines whether a copy of columns should be made when creating the DataFrame; by default, no copy is made.
+Read and parses a delimited file, materializing directly using the `sink` function.
 
-`CSV.read` supports the same keyword arguments as [`CSV.File`](@ref).
+`CSV.read` supports all the same keyword arguments as [`CSV.File`](@ref).
 """
-function read(source; copycols::Bool=false, kwargs...)
-    @warn "`CSV.read(input; kw...)` is deprecated in favor of `DataFrame!(CSV.File(input; kw...))`"
-    DataFrame(CSV.File(source; kwargs...), copycols=copycols)
+function read(source, sink=nothing; copycols::Bool=false, kwargs...)
+    if sink === nothing
+        @warn "`CSV.read(input; kw...)` is deprecated in favor of `CSV.read(input, DataFrame; kw...)"
+        sink = DataFrame
+    end
+    Tables.CopiedColumns(CSV.File(source; kwargs...)) |> sink
 end
-
-DataFrames.DataFrame(f::CSV.File; copycols::Bool=true) = DataFrame(getcolumns(f), getnames(f); copycols=copycols)
 
 include("precompile.jl")
 _precompile_()


src/CSV.jl
```diff
@@ -30,19 +30,20 @@ include("rows.jl")
...
-`CSV.read` supports the same keyword arguments as [`CSV.File`](@ref).
+`CSV.read` supports all the same keyword arguments as [`CSV.File`](@ref).
 """
-function read(source; copycols::Bool=false, kwargs...)
-    @warn "`CSV.read(input; kw...)` is deprecated in favor of `DataFrame!(CSV.File(input; kw...))`"
-    DataFrame(CSV.File(source; kwargs...), copycols=copycols)
+function read(source, sink=nothing; copycols::Bool=false, kwargs...)
+    if sink === nothing
+        @warn "`CSV.read(input; kw...)` is deprecated in favor of `CSV.read(input, DataFrame; kw...)"
+        sink = DataFrame
```
bkamins, line 42: I understand that with the current design when creating `DataFrame` we do not need to copy columns - right?
```suggestion
        @warn "`CSV.read(input; kw...)` is deprecated in favor of `CSV.read(input, DataFrame!; kw...)"
        sink = DataFrame!
```

bkamins:
>  what to do with `CSV.read(file)` when no sink argument is passed

I would disallow it

> one's a bit awkward as a form of "blessed" type piracy, but could also be useful for users as convenience

I think it is better to require specifying sink.

piever:
> The only question left in my mind is, with the 1.0 release, what to do with `CSV.read(file)` when no sink argument is passed. 

From the point of view of the casual user who has never heard of DataFrames and just wants to read a `.csv` file into something they can understand, I imagine it would be useful to default to a simple sink made of simple Base julia types, like `Tables.columntable`.

Unfortunately, this could be problematic for wide data (hopefully future julia versions will be able to deal with big named tuples). Maybe it's safer to disallow for now, and only decide on this later (it can be a post 1.0 decision if on 1.0 it errors).

bkamins:
> it can be a post 1.0 decision if on 1.0 it errors

This is one of the reasons I prefer throwing an error for now (@nalimilan :tm:)

src/CSV.jl
piever: I thought the point of `CopiedColumns` was to signal to the sink that copying is not necessary (that is, nobody "owns" the columns), so `DataFrame(c::CopiedColumns)` can be defined to be non-copying. If I understand the logic of `fromcolumns`, chances are that should already be the case.

quinnj: @piever is correct; DataFrames.jl already specializes `fromcolumns` for `CopiedColumns` to ensure no additional copies are made; so in this case, it doesn't matter if the user passes `DataFrame` or `DataFrame!`.

bkamins: Ah - right. This is nice indeed.

JeffBezanson:
:+1: Very in favor of this. I also think it would be fine to make `DataFrame` the default sink. The possibility of different types of tables is not *SO* important that every user must be forced to think about it from the first time they read a csv file in Julia.

JeffBezanson:
Ah, I understand you want to remove the DataFrames dependency. I think https://github.com/JuliaData/DataFrames.jl/issues/1764 is the way to go --- if that happens then maybe we can add back a default sink.

PR #704, opened 2020-07-31
Use depwarn for deprecations

omus:
Changes deprecation warnings to use `Base.depwarn` instead of `@warn`. The primary advantage is that you can use `--depwarn=no` to suppress deprecations or `--depwarn=error` to get a full stack trace.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -38,7 +38,7 @@ Read and parses a delimited file, materializing directly using the `sink` functi
 """
 function read(source, sink=nothing; copycols::Bool=false, kwargs...)
     if sink === nothing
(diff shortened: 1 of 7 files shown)


tomyun:
I believe we should have used `Base.depwarn(...; force=true)` here, at least for some critical changes like `CSV.read()`. As most users running Julia under the default `--depwarn=no`, there has been a slim chance seeing these deprecation warnings beforehand.

quinnj:
Thanks for bringing this up @tomyun, I agree. I've made a PR to make a new 0.7 patch release to set force=true for the depwarns; this will allow packages on the 0.7 release to see the depwarns before updating. https://github.com/JuliaData/CSV.jl/pull/782

Issue #708, opened 2020-08-01
read zipfile


kafisatz:
The example (from the Readme) below throws a warning.
It is not apparent to me what the correct way to do this would be (as ZipFile has no documentation)

```
# read file from zip archive
z = ZipFile.Reader("a.zip") # or "a2.zip"

# identify the right file in zip
a_file_in_zip = filter(x->x.name == "a.csv", z.files)[1]

a_copy = CSV.File(a_file_in_zip) |> DataFrame

┌ Warning: `CSV.File` or `CSV.Rows` with `ZipFile.ReadableFile` object is deprecated; pass a filename, `IOBuffer`, or byte buffer directly (via `read(x)`)
└ @ CSV C:\Users\me\.julia\packages\CSV\vohbW\src\utils.jl:239
```

quinnj:
I should just be `a_copy = CSV.File(read(a_file_in_zip)) |> DataFrame`

kafisatz:
thanks. I was about to try this.
SORRY. the warning actually says so.

PR #709, opened 2020-08-01
as per https://github.com/JuliaData/CSV.jl/issues/708

kafisatz:
as per https://github.com/JuliaData/CSV.jl/issues/708


quinnj:
Thanks!

PR #715, opened 2020-08-09
Add canonical url to all generated docs pages to promote latest docs in SEO

ianshmean:
People have been seeing old CSV docs versions in google search results, so this provides the canonical url to all docs versions, to promote the latest docs


mortenpi:
LGTM

ianshmean:
@mortenpi is there an additional setting that needs to be set so that new docs have the canonical link too?

quinnj:
Thanks for this!

nalimilan:
Apparently this wasn't enough to stop Google (even after #738): a search for "CSV.jl" in private mode still returns links to 0.1.1 and 0.2.5 in second and third position (the first link is the last version). Yet these pages have been indexed on September 18 and their cached version does include the canonical URL. Any ideas?

PR #716, opened 2020-08-09
make ignoreemptylines=true the default

### LINKED ISSUE #638: ignoreemptylines should default to true
This is related to #621 , right now `ignoreemptylines`'s default is `false`. I think it should default to `true`. 

I cannot imagine that the majority of people that want to parse a csv file intend to have missing in all the columns by leaving a line empty. Like, how would that even come to pass (as in, what program encodes `missing` in all columns by leaving an empty line)? 
Empty lines should be flagged as they are now (there is a warning about it), but they should also be ignored since the majority of users will want to filter those away since they stem from some error (such as shown in #621 or some other source) and not an intent.

Moelf:
fix #638


test/testfiles.jl
```diff
@@ -1,4 +1,10 @@
 function testfile(file, kwargs, expected_sz, expected_sch, testfunc; dir=dir)
+    kwargs = if :ignoreemptylines ∉ kwargs
+        (ignoreemptylines=false, kwargs...)
```
quinnj, line 3: Hmmm, this is the only case that makes me a little uncomfortable. I know it's a bit of a pain, but I'd rather go through all the test files and add `ignoreemptylines=false` manually if necessary; my guess is it won't affect the majority of cases.

Moelf:
@quinnj sorry for the delay, I've fixed as you suggested (i.e now each file test needs `ignoreemptylines=false` is set manually), but

the one that doesn't pass is because I can't get it to throw an error no matter what
```
julia> CSV.File(IOBuffer("x\n1\n2\n3\n#4"), ignoreemptylines=true)
4-element CSV.File{false}:
 CSV.Row: (x = "1",)
 CSV.Row: (x = "2",)
 CSV.Row: (x = "3",)
 CSV.Row: (x = "#4",)

julia> CSV.File(IOBuffer("x\n1\n2\n3\n#4"), ignoreemptylines=false)
4-element CSV.File{false}:
 CSV.Row: (x = "1",)
 CSV.Row: (x = "2",)
 CSV.Row: (x = "3",)
 CSV.Row: (x = "#4",)
```

quinnj:
I couldn't push to @Moelf 's branch, so I pushed the test fix here. This looks good to me now.

Moelf:
oops, forgot to check "Allow edits by maintainers", thanks!

PR #728, opened 2020-08-17
Use Logging.warn to output warning messages

gregod:
Using println for the warning messages in src/file.jl outputs directly to stdout. This creates issues when using a julia script in an unix pipe writing data to stdout. 

This pull request replaces these calls with the warn method from the stdlib logging framework. They correctly attach to stderr instead, preventing the mix of messages and data.


quinnj:
@oxinabox, @bkamins, does this look good to you two? (or anyone else have an opinion here?). Seems fine to me.

PR #731, opened 2020-08-18
Warn if user passes threaded=true, but there aren't enough rows

### LINKED ISSUE #723: Problem with threading on short files
Start Julia with 8 threads and run the following:
```
julia> CSV.write("test.csv", DataFrame(rand(4,3)))
"test.csv"

julia> CSV.File("test.csv")
4-element CSV.File{false}:
 CSV.Row: (x1 = 0.9582551431689226, x2 = 0.2547655765772492, x3 = 0.0679930619968423)
 CSV.Row: (x1 = 0.9093686896485851, x2 = 0.3288467627397862, x3 = 0.4685939563909609)
 CSV.Row: (x1 = 0.2679953837521867, x2 = 0.016865954623241697, x3 = 0.45856551627805553)
 CSV.Row: (x1 = 0.08766669790847903, x2 = 0.3104793946025721, x3 = 0.15814539954321094)

julia> CSV.File("test.csv", threaded=true)
ERROR: 2; something went wrong trying to determine row positions for multithreading; often it's a mismatch in the # of columns expected in the file vs. actual; pass `threaded=false` to avoid using multithreaded parsing; if you think you're seeing this error incorrectly, it'd be very helpful if you could open an issue at https://github.com/JuliaData/CSV.jl/issues so package authors can investigate
```

quinnj:
Fixes #723. The issue is that for threaded parsing, we need a minimum of
5 rows per task to correctly check that each threaded chunk starts at a
proper row start position; if we don't have at least 5 rows per task,
then we'll now issue a warning and set `threaded = false`.

### DIFF
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -526,6 +526,11 @@ f = CSV.File(transcode(GzipDecompressor, Mmap.mmap(joinpath(dir, "randoms.csv.gz
 @test length(f) == 70000
 @test eltype(f.first) == String
 
+# 723
+f = CSV.File(IOBuffer("col1,col2,col3\n1.0,2.0,3.0\n1.0,2.0,3.0\n1.0,2.0,3.0\n1.0,2.0,3.0\n"); threaded=true)
+@test length(f) == 4
+@test f isa CSV.File{false}
+
 # 706
 f = CSV.File(IOBuffer("a,b\n1,2"); types = Dict{Symbol,Type}(
                :a => CategoricalValue{String,UInt32},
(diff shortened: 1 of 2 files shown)


test/basics.jl
```diff
@@ -526,6 +526,11 @@ f = CSV.File(transcode(GzipDecompressor, Mmap.mmap(joinpath(dir, "randoms.csv.gz
 @test length(f) == 70000
 @test eltype(f.first) == String
 
+# 723
+f = CSV.File(IOBuffer("col1,col2,col3\n1.0,2.0,3.0\n1.0,2.0,3.0\n1.0,2.0,3.0\n1.0,2.0,3.0\n"); threaded=true)
```
bkamins, line 530: do you have CI set up to use multiple threads?

quinnj: Yes; we start julia with 1 and 2 threads.

PR #738, opened 2020-09-08
Add canonical URL to older versions of manual

nalimilan:
https://github.com/JuliaData/CSV.jl/pull/715 missed some releases. Spotted [on Discourse](https://discourse.julialang.org/t/csv-read-fails-with-keyword-arguments/46231).

I wasn't sure whether `{% if canonical_url %}` was defined or not, so I overwrote it with an explicit line.


quinnj (approved):
Sounds good to me!

PR #756, opened 2020-10-19
fix an error with writing Any[(a=1,)]

bkamins:


staticfloat:
Confirmed that this fixes my issue on slack.  Thanks @bkamins!

bkamins:
Ah - thank you for the fix. I thought it was handled earlier.

PR #766, opened 2020-10-31
Remove CategoricalArrays/DataFrames dependencies

quinnj:
These have been deprecated for a while now, so in preparation for the
0.8 release, we remove support for the `categorical` keyword argument,
producing `CategoricalArray` columns when the column type is given as
`CategoricalValue`, and `CSV.read` without an explicit sink argument.

I'd like to test a bit more what happens if you provide
CategoricalArray to `CSV.write`, or `CategoricalValue` as a column type.

For reference, this improves the loading time of CSV.jl on my machine from 1.05s, to 0.65s, around ~45% improvement


test/runtests.jl
```diff
@@ -22,25 +22,6 @@ end # @testset "CSV.File"
 
 include("write.jl")
 
-@testset "CategoricalArray levels (including ordering)" begin
-    f = CSV.File(IOBuffer("X\nb\nc\na\nc"), types=[CategoricalValue{String, UInt32}])
-    v = f.X[1]
-    @test v == "b"
-    @test levels(v.pool) == ["a", "b", "c"]
```
nalimilan, line 29: Looks like this test could be kept?

nalimilan:
How about testing `types=[CategoricalValue{String, UInt32}]`? Is that supposed to work?

quinnj:
Hmmm, yeah, that's currently giving:
```julia
julia> using CSV, CategoricalArrays
[ Info: Precompiling CSV [336ed68f-0bac-5ca0-87d4-7b16caf5d00b]

julia> f = CSV.File(IOBuffer("X\nb\nc\na\nc"), types=[CategoricalValue{String, UInt32}])
ERROR: MethodError: no method matching parse(::Type{CategoricalValue{String, UInt32}}, ::String)
Closest candidates are:
  parse(::Type{Sockets.IPAddr}, ::AbstractString) at /Users/jacobquinn/julia/usr/share/julia/stdlib/v1.6/Sockets/src/IPAddr.jl:246
  parse(::Type{T}, ::AbstractString; base) where T<:Integer at parse.jl:240
  parse(::Type{T}, ::AbstractString; kwargs...) where T<:Real at parse.jl:379
  ...
Stacktrace:
 [1] xparse
   @ ~/.julia/dev/Parsers/src/Parsers.jl:754 [inlined]
 [2] parsevalue!(#unused#::Type{CategoricalValue{String, UInt32}}, flag::UInt8, column::SentinelArrays.SentinelVector{CategoricalValue{String, UInt32}, UndefInitializer, Missing, Vector{CategoricalValue{String, UInt32}}}, columns::Vector{AbstractVector{T} where T}, buf::Vector{UInt8}, pos::Int64, len::Int64, options::Parsers.Options{false, true, true, false, Missing, UInt8, Nothing}, row::Int64, rowoffset::Int64, col::Int64, types::Vector{Type}, flags::Vector{UInt8})
   @ CSV ~/.julia/dev/CSV/src/file.jl:907
```

We need to decide what to do here:
* just tell people that isn't supported anymore; maybe try to special-case `CategoricalValue` specifically and throw helpful error
* try to support it somehow; maybe through use of `DataAPI.defaultarray`, but I think we'd also need a way for `CategoricalValue` to tell us what kind of "wrapped" type it has

nalimilan:
OK. Probably not the end of the world.

`DataAPI.defaultarray` is already implemented by CategoricalArrays, but `parse` isn't and I'm not sure it should. Actually currently you can't construct a `CategoricalValue` in isolation, you can only get it via its parent pool or array: anyway it would be super slow to create a new `CategoricalValue` with a new pool for each entry. So I guess the only solution would be to assign the value to the array and let the conversion happen at that point. You're right that it would require knowing the wrapped type -- or fall back to just assignig the string without parsing it, which would work only for `CategoricalValue{String}` (but it's the most common type).

PR #775, opened 2020-11-15
Improve error message for CSV.read(source)

nalimilan:
Cf. https://discourse.julialang.org/t/csv-read-error-provide-a-valid-sink-argument/50157/6.


quinnj (approved):
Thanks

PR #786, opened 2020-11-21
Avoid ambiguity between user-provided UInt64 column type and PosLen

### LINKED ISSUE #785: UInt64 column type is parsed incorrectly as strings
Using CSV.jl v0.8.0 in julia 1.5.2
If I try to specify UInt64 as a column type, and load integer data from a csv-file, the resulting data has type String.
Below a minimum example:

```julia
using CSV
csv = """1,1,1,1,1,1
2,2,2,2,2,2
3,3,3,3,3,3"""
csvfile = CSV.File(IOBuffer(csv); header=0, types=[UInt64,Int64,UInt128,Int128,UInt32,Int32])
```
Only the first column has a type String and not the expected UInt64. The other columns are correctly parsed into their corresponding types.
```julia
typeof(csvfile[1].Column1) == UInt64
````
returns false.

quinnj:
Fixes #785. The issue here is that we were overloading `allocate(T,
len)` for various types, including our custom `PosLen` type when strings
are just parsed as a position and length. The problem is `PosLen` is
defined as `const PosLen = UInt64`, which means for a user-provided
`UInt64` type, it was allocating a plain `Vector{UInt64}` instead of the
expected `SentinelVector{UInt64}`. We were already special-casing the
allocation of the `PosLen` column, so the fix is pretty easy; create a
new `allocateposlen` instead of relying on the `allocate` overload.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -171,7 +171,8 @@ const SmallIntegers = Union{Int8, UInt8, Int16, UInt16, Int32, UInt32}
 function allocate(rowsguess, ncols, types, flags, refs)
     columns = Vector{AbstractVector}(undef, ncols)
     for i = 1:ncols
-        @inbounds columns[i] = allocate(lazystrings(flags[i]) && (types[i] === String || types[i] === Union{String, Missing}) ? PosLen : types[i], rowsguess)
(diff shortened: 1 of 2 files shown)


quinnj:
Closing in favor of https://github.com/JuliaData/CSV.jl/pull/787

PR #823, opened 2021-04-08
Don't make PointerString a subtype of AbstractString

andreasnoack:
...to avoid invalidating base methods. This reduces the number of invalidated methods from 953 to 15 when loading `CSV`.


quinnj:
Thanks @andreasnoack!

Issue #834, opened 2021-05-09
Reading Transposed CSV with unequal rows reads only minimum size of data


shakisparki:
### Data (test.txt)
a, 0.1, 0.2, 0.3
b, 0.4

### Code
`julia> f = CSV.File(read("test.txt"), transpose=true, normalizenames=true)`
`1-element CSV.File{false}:`
` CSV.Row: (a = 0.1, b = 0.4)`

### Issue
 Output ignores remaining data in first row instead of reading all data for row 1, and row 2 while reporting "missing" data in row 2.

quinnj:
Fixed in #837

Issue #810, opened 2021-02-22
Maybe allow mixing string/symbol/int in select/drop options


quinnj:
https://discourse.julialang.org/t/csv-jl-mixing-symbol-and-column-number-in-select-option/55450

quinnj:
Ok, this will be implemented in #837

PR #837, opened 2021-05-23
CSV parsing internals refactoring

quinnj:
This PR is a large refactoring of CSV.jl internals to simplify code, fix
a few bugs, and allow several desired improvements. Here's a list of
goals for the refactor:
  * Switch the `lazystrings` keyword argument to a more general
  `stringtype` argument, allowing users to pass
  `stringtype=PosLenString` to achieve the same effect
  `lazystrings=true` has previously provided; but in addition to just
  return string columns as a lazy string type, it allows this new lazy
  string type (`PosLenString`), so be used in pooled columns as well
  * Change the header.jl file to context.jl and `Header` struct to
  `Context`; it's a more descriptive name as it holds all the variables
  and initial computation done before any real parsing work begins, so
  it provides "context" to parsing
  * Updated multithreaded parsing to do "type sampling" before parsing
  begins; we were already scanning a number of rows per chunk when
  identifying where each chunk should start parsing; we now merge that
  operation with also tracking the values parsed while identifying row
  starts, and use these sample values to figure out a column's initial
  type, build up pooling column refpools, and determine other properties
  like max string size
  * Includes a fix for https://github.com/JuliaData/CSV.jl/issues/834,
  since I was in the middle of a large refactoring anyway and it wasn't
  too hard to add while mucking around w/ everything
  * Big update to how pooling is done; I've tried to explain the new
  approach in great detail in the src/README.md document, which I plan
  on having as a sort of "dev docs" going forward
  * Getting rid of the internal "flags" bitmask in favor of a structured
  `Column` type that holds current type information, pooled-ness, parsed
  values, plus individual fields for things the flags were previously
  tracking. Overall, this helps simplify quite a bit of code
  * Introduces new behavior where _extra columns_ detected while parsing
  will be _added_ to the returned final columnset, instead of warned
  about; this is something I've wanted to do for a long time, under the
  general rule of "don't discard present data"; if it's there, parse it.

Things I'd still like to do before merging:
  * Review codegen/performance; from a quick glance at a few files, it
  seems that pooling is slower; it also seems like `detect` has some
  dynamic dispatch causing allocations
  * Review keyword argument defaults for `CSV.File`/`CSV.read`; we're
  getting very close to tagging a 1.0, so I'd like to do a thorough
  review of parsing options and make sure we feel confident in the
  user-facing API
  * Docs/tests update


Project.toml
```diff
@@ -11,22 +11,22 @@ PooledArrays = "2dfb63ee-cc39-5dd5-95bd-886bf059d720"
...
 [compat]
 FilePathsBase = "0.6,0.7,0.8,0.9"
 Parsers = "1"
 PooledArrays = "0.5, 1.0"
 SentinelArrays = "1.2"
 Tables = "1.1"
-WeakRefStrings = "0.5,0.6"
-julia = "1"
+WeakRefStrings = "0.5,0.6,1"
+julia = "1.3"
```
bkamins, line 23: so CSV.jl will not work on Julia 1.0 after these changes?

src/CSV.jl
```diff
@@ -16,16 +22,22 @@ using PooledArrays
...
+export PosLenString
 
 struct Error <: Exception
     msg::String
 end
 
 Base.showerror(io::IO, e::Error) = println(io, e.msg)
 
+# constants
+const SINGLE_THREADED_POOL_DEFAULT = 0.75
```
bkamins, line 35: this is % of unique values? Why do you think for single thread it makes sense to make it as high as 75%?

src/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+By providing the `pool` keyword argument, users can control how this optimization will be applied to individual columns, or to all columns of the delimted text being read.
+
+Valid inputs for `pool` include:
+  * A `Bool`, `true` or `false`, which will apply to _all_ columns parsed; they either will _all_ be pooled, or _all_ not pooled
+  * A `Real`, which will be converted to `Float64`, which should be a value between `0.0` and `1.0`, to indicate the % cardinality threshold _under which_ a column will be pooled. e.g. by passing `pool=0.1`, if a column has less than 10% unique values, it will end up as a `PooledArray`, otherwise a normal array. Like the `Bool` argument, this will apply the same % threshold to _all_ columns
```
bkamins, line 15: maybe add a comment how this threshold is checked in the code (as I understand some heuristics is used here - right?)

src/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## The `pool` keyword argument
+
+CSV.jl provides a native integration with the [PooledArrays.jl](https://github.com/JuliaData/PooledArrays.jl/) package, which provides an array storage optimization by having a (hopefully) small pool of "exspensive" (big or heap-allocated, or whatever) values, along with a memory-efficient integer array of "refs" where each ref maps to one of the values in the pool. This is sometimes referred to as a "dictionary encoding" in various data formats. As an example, if you have a column with 1,000,000 elements, but only 10 unique string values, you can have a `Vector{String}` pool to store the 10 unique strings and give each a unique `UInt32` value, and a `Vector{UInt32}` "ref array" for the million elements, where each element just indexes into the pool to get the actual string value.
+
+By providing the `pool` keyword argument, users can control how this optimization will be applied to individual columns, or to all columns of the delimted text being read.
+
+Valid inputs for `pool` include:
+  * A `Bool`, `true` or `false`, which will apply to _all_ columns parsed; they either will _all_ be pooled, or _all_ not pooled
```
bkamins, line 14: do we think this is useful in practice? E.g. under these rules `Bool` column would be pooled.

src/README.md
```diff
@@ -0,0 +1,26 @@
...
+  * A `Real`, which will be converted to `Float64`, which should be a value between `0.0` and `1.0`, to indicate the % cardinality threshold _under which_ a column will be pooled. e.g. by passing `pool=0.1`, if a column has less than 10% unique values, it will end up as a `PooledArray`, otherwise a normal array. Like the `Bool` argument, this will apply the same % threshold to _all_ columns
+  * As a special-case, `NaN` (the `Float64` not-a-number value), is used to signal "default" pooling behavior. "Default" here means that only columns with `String` values will be considered for pooling, and then a different threshold will be considered based on whether a file is being parsed by multiple threads or not; for single-threaded parsing, the default threshold for pooling is `0.75`. For multi-threaded parsing, the default threshold is `0.25`. The reasoning here is that single-threaded parsing scenarios are almost always smaller files, where the relative cost of pooling is smaller to the benefits of a pooled column output. For multithreaded parsing, pooled reference values must be synchronized across threads, and it can be costly to "unpool" a column after parsing the entire file.
```
bkamins, line 16: `String` or just "string" (as I guess eventually `String` does not have to be used?)

src/README.md
```diff
@@ -0,0 +1,26 @@
...
+  * An `AbstractVector`, where the # of elements should/needs to match the # of columns in the dataset. Each element of the `pool` argument should be a `Bool` or `Real` indicating the pooling behavior for each specific column.
```
bkamins, line 17: why "should/needs" - I understand it is "must" (if it does not match we throw an error - right?)

src/README.md
bkamins, line 16: a small detail only `Float64` `NaN` is treated this way or `NaN` of any type (as checked by `isnan`?)?

src/README.md
```diff
@@ -0,0 +1,26 @@
...
+  * An `AbstractDict`, with keys as `String`s, `Symbol`s, or `Int`s referring to column names or indices, and values in the `AbstractDict` being `Bool` or `Real` to again signal how specific columns should be pooled
```
bkamins, line 18: Again - do you require `String` or `AbstractString`. Similarly - do you require `Int` or any `Union{Signed, Unsigned}`?

bkamins: also is it allowed to mix strings, `Symbol` and integers or not?

src/README.md
```diff
@@ -0,0 +1,26 @@
...
...n start with when parsing their local chunks. Two tricky implementation details involved with sampling and pooling are 1) if a columns values end up being promoted to a different type _while_ sampling or post-sampling while parsing, and 2) if the whole columnset parsed is widened. For the first case, we take all sampled values and promote to the "widest" type, then when building a potential refpool, only consider values that "are" (i.e. `val isa type`) of the promoted type. That may seem like the obvious course of action, but consider that we may detect a value like `2021-01-01` as a `Date` object, but the column type is promoted to `String`; in that case, the `Date(2021, 1, 1)` object parsed _will not_ be in the initial refpool, since `!(Date(2021, 1, 1) isa String)`. For the 2nd tricky case, the columnset isn't widened while type sampling, so "extra" columns are just ignored. The columnset _will_ be widened by each local parsing task that detects the extra columns, and those extra columns will be synchronized/promoted post-parsing as needed. These "widened" columns will only ever be pooled if the user passed `pool=true`, meaning _every_ column for the whole file should be pooled.
```
bkamins, line 23:
```suggestion
  * For multithreaded parsing, we decide if a column will be pooled or not from the type sampling stage; if a column has a `pool` value of `1.0`, it will _always_ be pooled (as requested), if it has `0.0` it will _not_ be pooled, if `0.0 < pool < 1.0` then we'll calculate whether or not it will be pooled from the sampled values. As noted above, a `pool` value of `NaN` will also be considered if a column had `String` values sampled and meets the multithreaded threshold. Currently, we'll sample `lines_to_check * tasks` values per column, which are both configurable via keyword arguments, with defaults `lines_to_check=30` and `tasks=Threads.nthreads()`. From those samples, we'll calculate the # of unique values divided by the total # of values sampled and compare it with the `pool` value to determine whether the column will be ultimately pooled. The ramification here is that we may "get it wrong" in two different ways: 1) based on the values sampled, we may determine that a column _will_ be pooled even though the total # of uniques values we'll parse will be over the `pool` threshold and 2) we may determine a column _shouldn't_ be pooled because of seemingly high cardinality, even though the total # of unique values for the column is ultimately _lower_ than teh `pool` threshold. The only way to do things perfectly is to check _all_ the values of the entire column, but in multithreaded parsing, that would be expensive relative to the statistically accurate sampling method. The other unique piece of multithreaded parsing is that we form a column's initial `refpool` field from the sampled values, which individual task-parsing columns will then start with when parsing their local chunks. Two tricky implementation details involved with sampling and pooling are 1) if a columns values end up being promoted to a different type _while_ sampling or post-sampling while parsing, and 2) if the whole columnset parsed is widened. For the first case, we take all sampled values and promote to the "widest" type, then when building a potential refpool, only consider values that "are" (i.e. `val isa type`) of the promoted type. That may seem like the obvious course of action, but consider that we may detect a value like `2021-01-01` as a `Date` object, but the column type is promoted to `String`; in that case, the `Date(2021, 1, 1)` object parsed _will not_ be in the initial refpool, since `!(Date(2021, 1, 1) isa String)`. For the 2nd tricky case, the columnset isn't widened while type sampling, so "extra" columns are just ignored. The columnset _will_ be widened by each local parsing task that detects the extra columns, and those extra columns will be synchronized/promoted post-parsing as needed. These "widened" columns will only ever be pooled if the user passed `pool=true`, meaning _every_ column for the whole file should be pooled.
```

bkamins: (typo was fixed)

bkamins: Maybe we should increase `lines_to_check` from 30 to e.g. something like 1000 as a default sampling threshold? This should not be that expensive. Right?

src/README.md
```diff
@@ -0,0 +1,26 @@
...
+  * In the single-threaded case, we take a simpler approach by pooling any column with `pool` value `0.0 < pool <= 1.0`, meaning even if we're not totally sure the column will be pooled, we'll pool it while parsing, then decide post-parsing whether the column should actually be pooled or unpooled.
```
bkamins, line 24: can you comment why a similar approach is not used for multi-threaded parsing? I.e. we could pool per thread and only afterwards - before merging pools - use some heuristics to decide if pools should be merged (this should be more accurate than sampling based decision).

I am not super convinced that we have to change it. In general I would put a recommendation somewhere that users should preferably manually signal if the column should be pooled or not by `Bool` argument.

src/chunks.jl
```diff
@@ -41,7 +41,7 @@ function Chunks(source;
     ignoreemptylines::Bool=true,
     threaded::Union{Bool, Nothing}=nothing,
     tasks::Integer=Threads.nthreads(),
-    lines_to_check::Integer=5,
+    lines_to_check::Integer=30,
```
bkamins, line 44: maybe 1000 as commented above?

bkamins:
So after this change there will be no `PooledString` option. Right? We should use `pool` everywhere.

From the description it was not clear to me what happens if the user chooses `pool=true` for some column but also `lazystrings=true`. What will be the type of the produced column if pooling is done?

Also how is `PosLenStringVector` different from the old `LazyStringVector`? I thought it would be more flexible (i.e. allow subsetting, resizing etc. as long as we are not trying to change the pool of strings used)

src/CSV.jl
quinnj: I based it off a conversation we had elsewhere, where because single-threaded parsing is typically w/ smaller files, we could try to pool more columns by default, since that tends to lead to optimizations in downstream workflows. I don't really have a strong opinion on what we make this or if we should just keep it consistent in both cases.

src/README.md
quinnj: Not really heuristics; just that multithreaded parsing will only use _sampled_ values + # unique to determine pooling vs. entire file (single-threaded).

src/README.md
quinnj: Yeah, perhaps we should still limit this case to `stringtype` columns; that would let us get rid of the `NaN` special-case, because we could change the default back to `pool=0.1` (or whatever threshold value we want).

src/README.md
quinnj: Correct. It will be any string type supported as a valid input to `stringtype` keyword argument.

src/README.md
quinnj: Well, this was a question I had while implementing the "column widening" feature. If a user provides `header` or `types`, or other keyword argument for each column, should we accept no matter what? And then when parsing data, emit warnings on too few columns if necessary? The use-case I have in mind is that if you have a header row w/ too few column names, but you know actual data has _more_ columns, you might provide `header`/`types`/`pool` keyword arguments for all the columns _in the data_, and then we ignore the header row in the data. So this would mean we accept any length `header`/`types` `AbstractVector` keyword arguments, trusting the user knows how many columns in the data. I'm not totally sure this is the right thing to do though, since user may get it wrong.

src/README.md
quinnj: We check w/ `isnan`

src/README.md
quinnj: There is open issue to allow mixing strings/symbols/integers: https://github.com/JuliaData/CSV.jl/issues/810

src/README.md
quinnj: Hmmm, that would be worth checking. We should keep it in ratio w/ default multithreading cell-count number.

src/README.md
quinnj: Yes, manual is always better. The reason we determine pooling during sampling is performance; it's a little more expensive to pool a column while parsing, because we have to call `hash` and do `Dict` lookup in addition to setting array value of ref code. So ideally, we only pay that extra cost when we know we will pool.

quinnj:
> From the description it was not clear to me what happens if the user chooses `pool=true` for some column but also `lazystrings=true`. What will be the type of the produced column if pooling is done?

`lazystrings=true` is now deprecated, and just sets `stringtype=PosLenString` internally. So type of pooled column will match whatever is passed as `stringtype=X` (to allow for short strings, `String`, `PosLenString`, etc.).

> Also how is PosLenStringVector different from the old LazyStringVector? I thought it would be more flexible (i.e. allow subsetting, resizing etc. as long as we are not trying to change the pool of strings used)

`PosLenStringVector` is very similar to `LazyStringVector`, but will produce `PosLenString` elements on indexing, rather than `String`, thus avoiding full `String` allocations when using. It also supports a few limited operations via `similar`, `copyto!`, `getindex` w/ multiple inds that will produce `PosLenStringVector` instead of producing full `Vector{PosLenString}`.

src/CSV.jl
bkamins: OK - this perspective makes sense. Especially that - at least now - single and multi-threaded rules are not consistent in general (as one is exact and other is heuristic only)

src/README.md
bkamins: This is what I meant to add (but AFAICT it is explained later so this is OK; I was reading the PR linearly)

src/README.md
bkamins: I think pooling `<:Number` does not make much sense. I would restrict the default to `stringtype` columns. @nalimilan - any opinion here? Just to be sure: what column `eltype`s CSV.jl by default can create? I understand that apart from numbers and strings it is only dates. Right?

src/README.md
bkamins: OK - since this "should/needs" is because of just in case handling of column widening it is OK. What do you do now when you detect column widening? Emit a warning?

Project.toml
quinnj: I'll have to review; I think there was a dependency or two that needed 1.3

src/README.md
quinnj: Int64, Float64, Bool, Date, DateTime, Time, and `stringtype`. But we do allow passing custom types. I needed to generalize the pooling code to eventually support pooling short strings, and allowing any type naturally comes out of that. I agree it doesn't make sense to pool Bool and probably Float64, but there may be a use-case for Int64. I think it would be a good default to still only pool strings by default, but allow user to pool custom type/Int64 if requested.

src/README.md
quinnj: We don't emit a warning, but we probably should. Previous to this PR, we emit a warnign when too many columns are detected for row, now we parse them. I think it would probably be good to emit a warning though since it's rare behavior and probably unexpected.

src/README.md
bkamins: >  I think it would be a good default to still only pool strings by default, but allow user to pool custom type/Int64 if requested.

Agreed

src/CSV.jl
quinnj: What does @nalimilan think: too weird to have different pooling thresholds in single threaded vs. multithreaded? Should we just make them the same?

Project.toml
quinnj: Yes, with the SentinelArrays.jl ChainedVector code, there were a lot of updates/changes in array functions pre 1.3 that were a hassle to try and support, so SentinelArrays.jl is now 1.3

src/README.md
quinnj: Also note that the actual number of rows sampled for types is `lines_to_check * tasks`, where `tasks` defaults to `Threads.nthreads()`, so it will always be at least 60. But probably worth doing some performance tuning around what the best default is here; basically what's the highest we can go w/o really impacting total multithreaded parsing performance.

bkamins: I was aware that the number is per-thread. I just assumed that doing 1000 would be still fast enough.

bkamins: Fixed proportion makes sense, but I would think that if this proportion is less than 1000, we could use this number (i.e. for small files it is OK to just check more as it is fast anyway)

src/CSV.jl
quinnj: Ok, that's fine by me; I'll update to have a single default pool threshold value. Also happy to include a bias correction if someone wants to work out the math there; I've kind of just loosely been relying on the fact that we'll do a "reasonable number" of samples per chunk, which are naturally equally distributed across the file, so assume it's fairly representative of the full file. As Bogumil suggested, we can also bump up the default `lines_to_check` to increase the strength of that assumption.

src/README.md
quinnj: I'm pretty against throwing errors once we've started parsing, except for the one fatal case where a cell is not quoted properly. In my experience, and from what I've heard from others, it's really annoying to have parsing fail, especially in the case where it's a huge file, so it's parsing for a long time, then gets to the end and there happens to be a junk row that causes the whole parsing to fail. I'd be fine though with adding a "don't widen" keyword arg if we find out people actually _don't_ want this extra data parsed.

src/README.md
quinnj: I mean, we kind of give you the ability to check fixed proportion by configuring `lines_to_check`. The number of rows in the file is always a best guess until we're actually done parsing.

src/README.md
bkamins: I can understand this, therefore if you prefer this I would vote for three valued logic as proposed above:
* `nothing`: warning (the default)
* `true` or `false` - accept or throw an error

All three behaviors might be expected by users in different use cases.

src/CSV.jl
bkamins: > bias correction

We cannot reliably do this because we are not doing a random sampling but sequential sampling.
E.g. you can have the following data `repeat(1:10^6, 100)`. Then for sure you will not be able to detect if you should pool correctly no matter what bias correction you do (you will think all is unique while it is not).

Exactly because of this effect I voted for increasing the threshold to 1000. The reason is that having this value at least reduces the risks associated with the fact that we are not sampling randomly but sequentially.

nalimilan: Right, so we have to opposite biases:
- due to sequential sampling, values are more likely to be equal than in the full data -> we underestimate the proportion of unique values
- due to relying on a sample to estimate the proportion for the full data, values are more likely to be unique than in the full data -> we overestimate the proportion of unique value

AFAICT both biases can be somewhat limited by sampling a relatively large proportion of rows. Using at least 1000 rows sounds like a good idea, but I even suggest using a proportion of rows by default, so that more rows are sampled for large datasets, which prevents the bias from increasing. For example, for 100M rows, you may get 1000 unique values in a sample of 1000 rows, but there may actually be 10000 unique values and the data is sorted in some way (and in that case we should pool).

bkamins: > but I even suggest using a proportion of rows by default

I agree, but I understand that @quinnj is not able to know number of rows upfront. Normally I would suggest something like e.g. `maximum(1000, div(nrow(file), 1000))` but I understand this is not possible.

bkamins:
On:
```
(DF_DEV) pkg> st
      Status `C:\WORK\dev\DF_DEV\Project.toml`
  [69666777] Arrow v1.5.0
  [6e4b80f9] BenchmarkTools v1.0.0
  [336ed68f] CSV v0.8.5 `https://github.com/JuliaData/CSV.jl.git#jq/internalsrefactor`
  [324d7699] CategoricalArrays v0.10.0
  [a93c6f00] DataFrames v1.1.1 `C:\Users\bogum\.julia\dev\DataFrames`
  [2dfb63ee] PooledArrays v1.2.1 `https://github.com/JuliaData/PooledArrays.jl.git#main`
  [ea10d353] WeakRefStrings v0.6.2 `https://github.com/JuliaData/WeakRefStrings.jl.git#jq/inlinestrings`
```
I do
```
df = DataFrame([[randstring(i) for _ in 1:1000] for i in 1:20], :auto)
CSV.write("test.csv", df)
```
and then I get on Windows:
```
julia> df2 = CSV.read("test.csv", DataFrame)
ERROR: MethodError: no method matching (PooledArrays.PooledArray{WeakRefStrings.InlineString1, R, N, RA} where {R<:Integer, N, RA})(::PooledArrays.RefArray{Vector{UInt32}}, ::Dict{Union{Missing, WeakRefStrings.InlineString1}, UInt32})
Stacktrace:
 [1] makepooled!(col::CSV.Column)
```
Which seems to signal that you are using some branch of PooledArrays.jl. Which one should be used?

bkamins:
Ah PooledArrays.jl#bkamins-patch-1 is required. Now it works.

bkamins:
Below I report on the case where we have a regression.

The benchmarks in https://github.com/JuliaData/Arrow.jl/issues/188 for `N=10^7` are (I have used `pool=false` for now as the current state of how we use PooledArrays.jl should be changed):

this set of PRs:
```
julia> GC.gc(false); @time innerjoin(x, small, on = :id1, makeunique=true, matchmissing=:equal);
  0.402371 seconds (328 allocations: 849.167 MiB)

julia> GC.gc(false); @time leftjoin(x, medium, on = :id2, makeunique=true, matchmissing=:equal);
  0.567584 seconds (433 allocations: 1.095 GiB, 2.45% gc time)

julia> GC.gc(false); @time innerjoin(x, big, on = :id3, makeunique=true, matchmissing=:equal);
  3.729259 seconds (371 allocations: 1.237 GiB, 0.45% gc time)
```

current released versions:
```
julia> GC.gc(false); @time innerjoin(x, small, on = :id1, makeunique=true, matchmissing=:equal);
  0.417128 seconds (328 allocations: 780.502 MiB)

julia> GC.gc(false); @time leftjoin(x, medium, on = :id2, makeunique=true, matchmissing=:equal);
  0.547001 seconds (433 allocations: 1.002 GiB)

julia> GC.gc(false); @time innerjoin(x, big, on = :id3, makeunique=true, matchmissing=:equal);
  3.306644 seconds (371 allocations: 1.103 GiB)
```

These tests mostly check copying of `InlineString`. As you can see we have a small regression in `innerjoin` on `big`. The reason is (at least this is what I think) that we have there:
1. one column with `InlineString15`
2. two columns with `InlineString7`

If we are not making operations on these strings then these columns are more expensive to copy than just `Vector{String}` because they have a larger memory footprint (instead of pointers we store data which is larger).

This means that in the examples above (where the data is small enough not to trigger GC) using `String` is faster.

bkamins:
I think that given the amount of changes introduced it is essential not only to update docstrings, but also to prepare a list of examples of common scenarios the user might want to use (e.g. I am currently unsure what we would have pass in H2O benchmarks to make sure all numeric columns are loaded normally and all string columns are pooled independent on the count of unique values and loaded as inline strings).

quinnj:
> to make sure all numeric columns are loaded normally and all string columns are pooled independent on the count of unique values and loaded as inline strings

Numeric columns will not be pooled automatically. The only way to pool a numeric column is to request it specifically like `pool=Dict(1 => true)`. Passing `pool=true` means all string, and only string, columns _will_ be pooled, regardless of cardinality. Currently, the default is `stringtype=InlineString`, which means columns with individual string sizes < 32 will be inline string types, while columns with larger string values will be `String`.

So all that to say: you shouldn't have to pass anything by default. Unless you really want to make sure all string columns are pooled, but the logic should be much better now to automatically choose pooling or not. But probably the only thing we need to pass is `CSV.read(file, DataFrame; pool=true)`.

But I agree that I want to have a much better "manual" kind of documentation that walks through the various configuration options.

bkamins:
Thank you! I have forgotten to add that we should also read in integers as `Int32`, the example current command is:
```
CSV.read(src_grp, DataFrame, types=[PooledString, PooledString, PooledString, Int32, Int32, Int32, Int32, Int32, Float64], threaded=false);
```

quinnj:
> I have forgotten to add that we should also read in integers as `Int32`

Ah, interesting. Is that just for memory savings? Or does it make processing faster too? It wouldn't be too hard to allow reading `Int32` columns by default and promote to Int64 if needed.

bkamins:
> Ah, interesting. Is that just for memory savings? Or does it make processing faster too?

Both - it uses less memory and is cheaper to copy (and both are relevant in the benchmarks).

> It wouldn't be too hard to allow reading `Int32` columns by default and promote to `Int64` if needed.

I think it is OK to read in `Int64` as default, I just need to know how in the new API we should pass information that we want integers to be read in as `Int32`.

dkaszynski:
> I think it is OK to read in `Int64` as default, I just need to know how in the new API we should pass information that we want integers to be read in as `Int32`.

I agree that `Int64` as a default make sense. However, it would be great if there was an option similar to Pandas `downcast`, where for integer columns you can get the smallest type that can hold the data in the given column. I often find it very useful when working with data that has very large volumes.

quinnj:
Ok, pushed code to support reading Int32 by default and promoting to Int64 if necessary.

bkamins:
> Ok, pushed code to support reading Int32 by default and promoting to Int64 if necessary.

I think it is also OK to use this approach, as `Int32` should be enough in most cases as a default.

@nalimilan - agreed?

nalimilan:
Makes sense, though I'm somewhat afraid that people might run into overflow issues without noticing...

If we consider that overflow isn't an issue, how about using `Int8` and `Int16` by default where possible too? In my field at least, most CSV files consist of categorical columns coded using integers, which are often below 10. So most columns would end up being `Int8`, using 4 times less memory. Or would that be too complex for users?

bkamins:
> Makes sense, though I'm somewhat afraid that people might run into overflow issues without noticing...

This was my second thought after I have written my comment. The problem is:
```
julia> x = Int32(2_000_000_000)
2000000000

julia> x+x
-294967296
```
is a completely "thinkable" scenario for normal data (while to get something like this with `Int64` you have to do heavy number crunching).

Maybe better keep to my original recommendation:
* read integers as `Int64` by default
* allow passing `downcast=true` that would fit the column into smallest possible type from the list `Int8`, `Int16`, `Int32`, and `Int64` (this would then fulfill @nalimilan's request while being explicit)

What do you think about it?

quinnj:
Yeah, that does sound better, though not entirely simple to implement. Let me take a look.

quinnj:
Ok, adding `downcast` wasn't too hard, and actually generalized the Int32 and other promotion code we already have pretty nicely. I haven't added docs for it yet as I'm starting to sketch out a broader plan for doc updates that are sorely needed. I'm considering doing more extensive sections for each keyword arg that would explain it's usage in detail, including examples, what it does/doesn't do, caveats, how it interacts with other keyword args, valid types of inputs, etc. I'm wondering if we need a brief list of keyword args like we have now, then more expanded docs? Or just do the expanded docs?

quinnj:
Ok, this PR has gotten a bit unwieldy and tests have been passing for a while (I'll investigate the nightly failure later, but it shouldn't be anything to worry about). Going to put up a new PR for cleaning up all the keyword arguments for CSV.File/CSV.read/CSV.rows/CSV.Chunks

Issue #860, opened 2021-07-30
Error when loading complex NaN's


DavidWellnitz:
I have a CSV File which contains complex values and some values are ```NaN + NaN*im```, which throws an error when I try to read it with type fixed to ```ComplexF64```. The underlying problem seems to be that ```parse(ComplexF64, "NaN + NaN*im")``` throws the error: ```ERROR: ArgumentError: cannot parse " NaN*" as Float64```. I tried to "solve" this by setting ```strict=false``` like:
```df = CSV.File(fname, header=header, type=ComplexF64, strict=false) |> DataFrame```
where ```fname``` is the location of my file and ```header``` is the header, but it throws the same error.

Am I using ```strict``` wrong? And is the behavior of ```parse``` intentional?

DavidWellnitz:
Update: Using ```missingstring= "NaN+NaN*im"``` instead of ```strict=false``` works, but it still feels like ```strict=false``` should also work.

quinnj:
Hmmm, yes, an interesting problem. For non-standard types, we currently rely on calling `Base.parse`, but on invalid values, that may throw an error that we're currently not catching. We should probably switch to using `Base.tryparse`, so that `strict=false` works as you mentioned. We could also consider adding parsing support in Parsers.jl for Complex types, which would make parsing them much faster.

quinnj:
Ok, this was fixed in https://github.com/JuliaData/Parsers.jl/pull/86

DavidWellnitz:
Thanks :)

PR #863, opened 2021-08-03
Improve error message in case input file doesn't exist

### LINKED ISSUE #859: Error when file is not found
The message in
https://github.com/JuliaData/CSV.jl/blob/0eaacb4c4d5787a5186dac8edf523ee9052db27f/src/context.jl#L179
is reported to be misleading by the users.

Maybe we could split the `isfile` condition to a separate test and if file is not found report that it was not found? (users think that "valid file" means that file exists but did not pass validation)

quinnj:
Attempt to fix #859.

### DIFF
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -6,6 +6,7 @@ struct CSVString
 end
 
 Base.parse(::Type{CSVString}, s::String) = CSVString(s)
+Base.tryparse(::Type{CSVString}, s::String) = CSVString(s)
 Base.zero(::Type{CSVString}) = CSVString("")
 
 struct Dec64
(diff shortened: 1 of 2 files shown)


bkamins (approved):
Looks good.

test/basics.jl
```diff
@@ -6,13 +6,15 @@ struct CSVString
 end
 
 Base.parse(::Type{CSVString}, s::String) = CSVString(s)
+Base.tryparse(::Type{CSVString}, s::String) = CSVString(s)
```
bkamins, line 9: why are these definitions needed?

quinnj: In the fix for https://github.com/JuliaData/Parsers.jl/pull/86 and #860, Parsers now requires `tryparse` to be defined instead of `Base.parse` in the custom type fallback.

PR #864, opened 2021-08-04
Add automatic support for reading gzipped csv files

### LINKED ISSUE #733: Slow reading of gzipped files
Following https://voxeu.org/article/which-programming-language-best-economic-research

The slow thing is:
```
 GZip.open("crsp_daily.gz","r") do io
    CSV.read(io, DataFrame)
end
```

@quinnj - do you think it is GZip.jl related or rather how CSV.jl and GZip.jl interact?

quinnj:
Also fixes #733 by using a performant method for processing gzipped
files. The motivation here is to make this more "automatic" for users so
they don't have to fumble around and question whether they should use
GZip.jl or CodecZlib.jl or Zlib.jl or whatever. We know which is the
fastest (CodecZlib.jl), and it's a mature, properly binary-built package
at this point that works cross platform, so I submit that it's a
non-risky dependency to take on for the benefit it provides.

In terms of implementation, we take the same approach as data.table
`fread` here by decompressing the gzip file to a temp file, then
mmapping it. We track the temp files generated in this way and clean
them up at the end of `CSV.File` (I need to check `CSV.Rows` as well),
and we can also rely on having the default of `cleanup=true` for
`tempname()`, which means the julia process itself will try and delete
all its generated temp files when it exits.

Alternative approaches involving reading chunks of a
`GzipDecompressorStream` or having a new `BufferedIOVector <:
AbstractVector{UInt8}` which would buffer reads from a
`GzipDecompressorStream` input were both considered, but at the end of
the day, CSV.jl is hard-coded in several places to work on the input as
a `Vector{UInt8}` explicitly. Parsers.jl is more flexible, but it would
take non-trivial work to allow CSV.jl the same flexibility. Another
factor to consider here is that if we allowed many different `buf` types
(other than just `Vector{UInt8}`), then we dig up the recompilation
problems all over again (that we literally just finished burying with
the removal of type parameters from `Parsers.Options`). That's probably
the biggest factor for me personally to take the `fread` approach of
using tempfiles, in addition to being very performant.

So follow ups before merging are:
  * Check that `CSV.Rows` can clean up tempfiles
  * We can be more strict in a few places of requiring `Vector{UInt8}`
  in CSV.jl
  * Update doc examples to not need `transcode` explicitly


oscardssmith:
I'm not sure how I feel about this. On the one hand, it's probably a usability win. On the other, it seems wrong conceptually. Partly, this feels wrong because it makes GZip special, and although it is probably the most common compression, there are others that are widely used and often better (ZSTD, LZ4, BZIP). 

Naively, the `BufferedIOVector` seems like the best solution as it would not have significant TTFP implications (I think), and seems very flexible.

ericphanson:
I don’t know how common this is, but I often work on systems that have more free ram than free disk space, and pull files from AWS s3 directly into ram without touching disk (this can theoretically be very performant when in-network although the Julia s3 access story isn’t fully there yet). In this scenario, the tempfile solution would not be a good one.

I also don’t use CSVs very often so it’s kind of a moot point, but I wanted to bring up this kind of data access pattern as something that could be relevant.

iamed2:
I think it would be convenient if this allowed streams and generic `read`ibles to be gzip. Use case I'm thinking of is a `.csv.gz` file on S3, passed as either the stream from `s3_get(...; return_stream=true)` (hits the IO code path) or passed as an S3Path (hits the non-mmappable-file code path).

While the latter case doesn't have any performance benefits relative to decompressing the file manually and passing in the bytes, it makes it so that reading a local `.csv.gz` path and a remote one from S3 function the same way.

nalimilan:
I'm also torn about this, as it feels problematic to special-case .gz files. At least, would it be possible to allow passing other stream types as a keyword argument to get the same level of convenience for other compression formats?

In terms of implementation, do you think it would be possible to use anonymous mmap so that no file is actually created if there's enough RAM? Then the OS could decide to use the swap only if needed.

src/utils.jl
```diff
@@ -148,26 +148,39 @@ consumeBOM(buf, pos) = (length(buf) >= 3 && buf[pos] == 0xef && buf[pos + 1] ==
...
-            return buf, 1, length(buf)
+            # if we can't mmap, try just buffering the whole thing into a tempfile byte vector
+            buf, tfile = buffer_to_tempfile(CodecZlib.TranscodingStreams.Noop(), x)
+            return buf, 1, length(buf), tfile
         end
     end
 end
 
+function buffer_to_tempfile(codec, x)
+    file = tempname()
```
nalimilan, line 176: FWIW, `tempname` isn't considered as perfectly secure due to possible race conditions. `mktemp` is better.

quinnj:
@oscardssmith, I think gzip is by far the most common compression format used with csv files; next would be .zip archives, typically in windows environments. I'd like to have a way to handle .zips as well in the future.

@ericphanson , thanks for the comment; that's very valid, particularly for production cases where you're running on ram-heavy/disk-lite ec2 instances or whatever. I've now added the `buffer_in_memory` keyword argument, which defaults to `false`, which can be set to `true` in order to do gzip decompressing in memory instead of via temporary file.

@iamed2, very good point; I've changed the logic around to ensure that any source that has the gzip format will be decompressed automatically.

@nalimilan thanks for the pointer on `mktemp`; I've changed it.

PR #869, opened 2021-08-18
Overhaul CSV.jl docs

quinnj:
It's been a long time coming. Changes include:

  * In the `index.md`, we have an opening "manual" section that discusses the high-level functionality of the package and gives overview of eac (`CSV.File`, `CSV.read`, `CSV.write`, etc.)
  * We also have larger sections for each supported keyword argument to `CSV.File`, with notes of compat w/ `CSV.Rows`/`CSV.Chunks`. Almost all they keyword argument sections also have links to examples in the `examples.md` file
  * The `examples.md` file has been restructured to be easily linkable from the `index.md` file, and includes more in-depth comments inline for each example, which should almost entirely be copy-pastable to a local REPL.
 
 I've tried to dump as much information as hopefully is useful for users between the keyword argument sections and corresponding examples.
 
 If there are pieces that aren't clear, please comment and let's get it cleaned up! This is going to be the basis for 1.0 docs, so we'd like them to be solid, stable, and very helpful. Specifically, some of the trickiest pieces to clarify are the interactions between keyword arguments (like how `pool` might affect `types`, for one example); if you know or can think of cases where it's unclear how one keyword argument interacts with another, just chime in and we'll get it sorted out.
 
 Thanks for any and all review!


docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+# to instead have the encoding conversion happen in memory, pass
+# `buffer_in_memory=true`; this can be faster, but obviously results
+# in more memory being used rather than disk via a temp file
+file = CSV.File(open("iso8859_encoded_file.csv", enc"ISO-8859-1"); buffer_in_memory=true)
+```
+
+### [Gzipped input](@id gzipped_input)
+
+```julia
+# assume I have csv text data compressed via gzip
```
bkamins, line 41: I think it would be useful to tell the users:
* what compression formats would be automatically handled
* what extensions are automatically detected (unless it is solely contents based format detection - and if this is the case also comment on this)

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+a |> CSV.write(f)
+close(z)
+
+# read file from zip archive
+z = ZipFile.Reader("a.zip") # or "a2.zip"
+
+# identify the right file in zip
+a_file_in_zip = filter(x->x.name == "a.csv", z.files)[1]
+
+a_copy = CSV.File(read(a_file_in_zip)) |> DataFrame
```
bkamins, line 119: probably `z` needs to be closed again (just for cleanness) - right?

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+# and concatenate the values for each column, separating rows with `_` character
+file = CSV.File(IOBuffer(data); header=[1, 2])
+```
+
+### [Normalizing column names](@id normalize_header)
+
+```julia
+using CSV
+
+# in this case, our data are single letters, with column names of "1", "2", and "3"
```
bkamins, line 204: maybe add a comment that normalizing is not required but it is optional.

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+4,5,6
+7,8,9
+totals: 12, 15, 18
+grand total: 45
+"""
+
+# by passing `footerskip=2`, we tell parsing to start the end of the data and
+# read 2 rows, ignoring their contents, then mark the ending position where
+# the normal parsing process should finish
+file = CSV.File(IOBuffer(data); footerskip=2)
```
bkamins, line 272: a natural question is how to read exactly "up to" N rows of data

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+# now examples of dropping, when we'd rather specify the column(s)
+# we'd like to drop/exclude from parsing
+file = CSV.File(file; drop=[2])
+file = CSV.File(file; drop=[:b])
+file = CSV.File(file; drop=["b"])
+file = CSV.File(file; drop=[false, true, false])
+file = CSV.File(file; drop=(i, nm) -> i == 2)
+```
+
+### [Limiting number of rows from data](@id limit_example)
```
bkamins, line 366: ah - I had this comment above - I would move this section there as it is a more natural flow.

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+# terminate early, avoiding the parsing of the rest of the data entirely
+file = CSV.File(IOBuffer(data); limit=3)
+```
+
+### [Specifying custom missing strings](@id missing_string_example)
+
+```julia
+using CSV
+
+# in this data, our first column has "missing" values coded with -999
```
bkamins, line 394: maybe discuss what is treated as missing by default.
also a particular case is that NOTHING should be treated as missing and the question is how to do it.

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+
+file = CSV.File(file; truestrings=["T", "TRUE"], falsestrings=["F", "FALSE"])
+```
+
+### [Matrix-like Data](@id matrix_example)
+
+```julia
+using CSV
+
+# This file contains a 3x3 identity matrix of `Float64`. By default, parsing will detect the delimiter and type, but we can also explicitly pass `delim= ' '` and `types=Float64`, which tells parsing to explicitly treat each column as `Float64`, without having to guess the type on its own.
```
bkamins, line 542: probably `ignorerepeated` should be also added here, as this typically happens in matrix data:
```
julia> rand(5,5)
5×5 Matrix{Float64}:
 0.885415   0.436992  0.966233  0.415127   0.595875
 0.0674271  0.586727  0.992303  0.236927   0.997092
 0.729953   0.720313  0.70169   0.922141   0.377983
 0.932993   0.257281  0.92218   0.0864256  0.0153523
 0.852068   0.591985  0.577147  0.966295   0.220887
```

docs/src/index.md
```diff
@@ -1,407 +1,258 @@
...
-## Key Functions
-```@docs
-CSV.File
-CSV.Chunks
-CSV.Rows
-CSV.write
-CSV.RowWriter
-```
+That's quite a bit! Let's boil down a TL;DR:
+  * Just want to read a delimited file and do basic stuff with data? Use `CSV.File(file)` or `CSV.read(file, DataFrame)`
```
bkamins, line 21: they are not equivalent - i.e. calling `DataFrame` on `CSV.File` output copies columns (which will also change their type)

docs/src/index.md
```diff
@@ -1,407 +1,258 @@
...
+
+## [`stringtype`](@id stringtype)
+
+An argument that controls the precise type of string columns. Supported values are `InlineString` (the default), `PosLenString`, or `String`. The various string types are aimed at being mostly transparent to most users. In certain workflows, however, it can be advantageous to be more specific. Here's a quick rundown of the possible options:
+
+  * `InlineString`: a set of fixed-width, stack-allocated primitive types. Can take memory pressure off the GC because they aren't reference types/on the heap. For very large files with string columns that have a fairly low variance in string length, this can provide much better GC interaction than `String`. When string length has a high variance, it can lead to lots of "wasted space", since an entire column will be promoted to the smallest InlineString type that fits the longest string value. For small strings, that can mean a lot of wasted space when they're promoted to a high fixed-width.
```
bkamins, line 227: maybe add a cross-reference where the user can learn more.

bkamins:
I am OOO for a week, with scattered access to Internet, so I gave some quick comments. Thank you!

quinnj:
> I am OOO for a week, with scattered access to Internet, so I gave some quick comments. Thank you!

Glad you were able to take a look! Thank you!

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+file = CSV.File(file; header=false)
+file = CSV.File(file; header=false, delim=' ', types=Float64)
+```
+
+### [Providing types](@id types_example)
+
+```julia
+using CSV
+
+# In this file, our 3rd column has an invalid value on the 2nd row `invalid`. Let's imagine we'd still like to treat it as an `Int` column, and ignore the `invalid` value. The syntax examples provide several ways we can tell parsing to treat the 3rd column as `Int`, by referring to column index `3`, or column name with `Symbol` or `String`. We can also provide an entire `Vector` of types for each column (and which needs to match the length of columns in the file). There are two additional keyword arguments that control parsing behavior; in the first 4 syntax examples, we would see a warning printed like `"warning: invalid Int64 value on row 2, column 3"`. In the fifth example, passing `silencewarnings=true` will suppress this warning printing. In the last syntax example, passing `strict=true` will result in an error being thrown during parsing.
```
rikhuijzer, line 558: Maybe, a nitpicking suggestion: Place each sentence on a newline to ease review of the Git diff in the future.

docs/src/examples.md
```diff
@@ -0,0 +1,609 @@
...
+
+# In this file, we have U.S. zipcodes in the first column that we'd rather not treat as `Int`, but parsing will detect it as such. In the first syntax example, we pass `typemap=Dict(Int => String)`, which tells parsing to treat any detected `Int` columns as `String` instead. In the second syntax example, we alternatively set the `zipcode` column type manually.
+data = """
+zipcode,score
+03494,9.9
+12345,6.7
+84044,3.4
+"""
+
+file = CSV.File(file; typemap=Dict(Int => String))
```
rikhuijzer, line 587: Nice! I'm using CSV for a year now and never knew this functionality existed

docs/src/index.md
```diff
@@ -1,407 +1,258 @@
...
... `Bool`, `Float64`, vector of `Bool` or `Float64`, or dict mapping column number/name to `Bool` or `Float64`. As a `Bool`, controls absolutely whether a column will be pooled or not; if passed as a single `Bool` argument like `pool=true`, then all string columns will be pooled, regardless of cardinality. When passed as a `Float64`, the value should be between `0.0` and `1.0` indicating the threshold under which the % of unique values found in the column will result in the column being pooled. For example, if `pool=0.1`, then all string columns with a unique value % less than 10% will be returned as `PooledArray`, while other string columns will be normal string vectors. As mentioned, when the `pool` argument is a single `Bool` or `Float64`, only string columns will be considered for pooling. When a vector or dict is provided, the pooling for any column can be provided as a `Bool` or `Float64`. Similar to the [types](@ref types) argument, providing a vector to `pool` should have an element for each column in the data input, while a dict argument can map column number/name to `Bool` or `Float64` for specific columns. Unspecified columns will not be pooled when the argument is a dict.
```
bkamins, line 214: If I understand correctly if single value for `pool` is used it is applied to string columns only and otherwise if it is a vector or dict to any columns. Maybe highlight it more?

Also in this description I do not see what is the default (maybe I missed it, but I cannot find it)

docs/src/examples.md
quinnj: good call; added.

docs/src/examples.md
quinnj: added

docs/src/examples.md
quinnj: added

docs/src/examples.md
quinnj: This is discussed in the `## [`missingstring`](@id missingstring)` section of index.md

docs/src/examples.md
quinnj: Good suggestion! Done.

docs/src/index.md
quinnj: Ah yes, I considered this section not-quite-done. I think it'd be useful for you, @nalimilan, and I (and anyone else interested) to discuss what the default pooling behavior will be. On current `main` branch, the default pool value is `0.25`, and it applies the same to single and multithreading. The multithreaded case has also been updated w/ the same behavior as single threaded: i.e. we'll parse the entire file, and post-parsing, we'll check the cardinality of _potentially_ pooled columns to see if they pass the threshold and then convert to regular array or pooled array as appropriate. The use-case I'm worried about given the current state is if you have a large file that uses multiple threads to parse, and you have a very high cardinality string column (like a string id). In that case, we would do roughly double the work parsing that id column, only to discard the pooled values at the end.

An alternative I've considered is making the default `pool=false`. So by default, no surprises or potential slowdowns. If you want pooled columns, you'd have to pass a pool value or `pool=true` explicitly.

quinnj:
> I just wanted to dump a big vanilla Julia matrix of numbers to a CSV file, so I found myself reading the docs. I couldn't find a single fully reproducible example (all of the examples seemed to presume some familiarity with Tables.jl) of how to dump an Array{Float64,2} to a CSV file anywhere. It seemed like something that would naturally belong in a quick start section...

@SamRodkey great suggestion! I just added a couple more examples explicitly for writing matrices; both to the high-level "start" and to the `CSV.write` docs specifically. These are exactly the kinds of suggestions that are the most helpful! Let me know if you have any more!

docs/src/index.md
bkamins: I am OK with both `pool=0.25.` and `pool=false` as a default. I think the problems users have with pooled columns should be resolved (like the `map` issue) on our side.

quinnj: Would it be too weird to have `pool=false` for multithreaded parsing and `pool=0.25` for single-threaded? Would we run into too many issues where the difference of array would be weird?

PR #872, opened 2021-08-20
Respect typemap in type promotion cases while parsing

### LINKED ISSUE #871: Keyword typemap has no effect
Observed behaviour:
```julia
julia> df = DataFrame(CSV.File("/home/jakob/Downloads/bnfilt.csv", typemap=Dict(Float64 => String)));

julia> typeof(df[:, "Prøve id"])
Vector{Float64} (alias for Array{Float64, 1})
```
I have verified this happens in CSV, not in DataFrames.
Unfortunately, any attempt of me to make a smaller example makes it work. I'm running with 1 thread in VSCode. Ref #797

quinnj:
Fixes #871. The issue here is that in our parsevalue! promotion logic,
we weren't considering whether the use may want to typemap a _promoted_
type to another type; in the original issue, a column started out as
detected as integer, then promoted to float, but the user wanted to
typemap any floats to strings. This fix just accounts for any
typemapping that may occur during promotion.

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -833,6 +833,11 @@ function parsevalue!(::Type{type}, buf, pos, len, row, rowoffset, i, col, ctx)::
                 if type === Int8 || type === Int16 || type === Int32 || type === Int64 || type === Int128
                     newT = _widen(type)
                     while newT !== nothing && !Parsers.ok(code)
+                        newT = get(ctx.typemap, newT, newT)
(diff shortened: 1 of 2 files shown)


shuofanzhang:
Still have this issue in Jupyter notebook, but no issue in VS Code. 
CSV v0.8.5
Jupyter Notebook 6.4.8

Issue #712, opened 2020-08-04
Bulk importing of CSV files


clibassi:
Ported from a discussion on Slack - was hoping there might be a way to combine multiple CSV files at once that will ultimately be appended to each other. Current options are multiple CSV.read() statements or: 

`df = CSV.read(firstfile)`
`for file in rest`
    `append!(df, CSV.File(file)`
`end`

quinnj:
Yeah, my thought here is we could have a `CSV.File(::Vector)` method that took a vector of filenames (or `IO` objects, or byte vectors), loop through each to do parsing, and append the results in a single CSV.File w/ `ChainedVector`s for each column, with chains being each file.

kmundnic:
Is having `CSV.File(::Vector)` necessary? I use dot notation for this:

```julia
df = vcat(CSV.read.(files, DataFrame)...)
```

rofinn:
That solution requires a dependency on DataFrames.jl. I think if CSV.jl handled this internally it could be a lot more efficient than splatting the results into a `vcat` call. That operation would be very slow if you had hundreds of csv files. Alternatively, if `Tables.Partitioner` was also a valid table then `CSV.File(::Vector)` could always return a lazy partitioner that you could then be materialized as whatever table type folks want?

quinnj:
DataFrames.jl has a specialized method on `vcat` that makes it pretty efficient. We already have `TableOperations.joinpartitions` that takes a valid `Tables.partitions` input and produces a "table", so you could do `df = Tables.Partitioner(CSV.File, myfiles) |> TableOperations.joinpartitions |> DataFrame`. `joinpartitions` operates lazily.

rofinn:
Okay, I'll open a PR to TableOperations.jl to add that to the README.md

rofinn:
Seems buggy?

```julia
  Got exception outside of a @test
  BoundsError: attempt to access 3-element Vector{AbstractVector{T} where T} at index [4]
  Stacktrace:
    [1] getindex
      @ ./array.jl:801 [inlined]
    [2] getcolumn
      @ ~/.julia/packages/CSV/Zl2ww/src/file.jl:52 [inlined]
    [3] getcolumn
      @ ~/.julia/packages/CSV/Zl2ww/src/file.jl:80 [inlined]
    [4] (::TableOperations.var"#21#24"{Vector{SentinelArrays.ChainedVector}, CSV.File{false}})(i::Int64)
[...]

PR #873, opened 2021-08-20
Allow passing vector of inputs to `CSV.File`/`CSV.read`

quinnj:
Implements #712. There's been some discussion back and forth on whether
we should allow something like this (passing a collection of inputs to
be all handled/processed together), or rely on more generic
infrastructure like `TableOperations.joinpartitions` with
`Tables.partitioner`. Upon reflecting and poking around a big, I think
it is worth having this builtin to `CSV.File`.

  1) It's not very much code, especially if we require the inputs to
  have identical schemas, as implemented in this PR. We could relax
  that requirement in the future and support various amounts of column
  promotion between files, but it seems conservative to just require
  strict schema requirements to start out.
  2) The proposed functionality in this PR automatically work with
  `CSV.read`. So if you do `CSV.read([file1, file2], DataFrame)`, you'll
  get a DataFrame with the results of `file1` and `file2` vertically
  concatenated. To get roughly the same result with generic functionality,
  you'd need to do something like `using Tables, TableOperations;
  Tables.partitioner(CSV.File, [file1, file2]) |>
  TableOperations.joinpartitions |> Tables.CopiedColumns |> DataFrame`,
  which is quite a mouthful.
  3) We can better control the use of threads with builtin
  functionality. In this PR implementation, we parse each input of the
  vector of inputs using a single thread, concurrently. So inputs are
  processed concurrently, but each one is only processed with a single
  thread. This has the advantage of avoiding nested `ChainedVector`s,
  which you would naively end up with in the previous point example
  using `TableOperations.joinpartitions`. In this PR, each input's
  columns will be plain vectors, and we use a single `ChainedVector` per
  column to vertically (and lazily) concatenate the results.


src/CSV.jl
```diff
@@ -21,6 +21,8 @@ using WeakRefStrings
 export PosLenString, InlineString
 # CodecZlib is used for unzipping gzip files
 using CodecZlib
+# FilePathsBase allows more structured file path types
+using FilePathsBase
```
rofinn, line 25: Sorry, why does this need to be a direct dependency? I don't see reference to a `AbstractPath` type below or the `Path` constructor`

src/file.jl
```diff
@@ -154,7 +101,64 @@ tbl = CSV.File(file) |> SQLite.load!(db, "sqlite_table")
...
+
+Tables.getcolumn(f::File, nm::Symbol) = getcolumn(f, nm).column
+Tables.getcolumn(f::File, i::Int) = getcolumn(f, i).column
+
+Base.@propagate_inbounds function Base.getindex(f::File, row::Int)
+    @boundscheck checkbounds(f, row)
+    return Row(getnames(f), getcolumns(f), getlookup(f), row)
+end
+
+function File(source::Union{Vector{UInt8}, IO, Cmd, AbstractString, AbstractPath};
```
rofinn, line 161: This seems weirdly specific and non-specific...? Maybe you could just special case the byte vector?

src/file.jl
```diff
@@ -960,3 +964,30 @@ end
...
+            files[i - 1] = File(sources[i]; kws...)
+        end
+    end
+    for i = 2:length(sources)
+        f2 = files[i - 1]
+        f.names == f2.names || throw(ArgumentError("column names don't match between delimited data inputs"))
+        f.types == f2.types || throw(ArgumentError("column types don't match between delimited data inputs"))
+        rows += f2.rows
+        for j = 1:length(f.columns)
+            append!(f.columns[j].column, f2.columns[j].column)
```
rofinn, line 989: I guess this will error if the first file has complete cases and subsequent files don't cause our initial `ChainedVector{T}` won't support subsequent columns of type `Union{T, Missing}`? Is the solution to pass in a full `typemap` for all possible types of `T => Union{T, Missing}`... cause that seems a bit verbose?

rofinn (commented):
Sorry, just got around to looking at this.

src/file.jl
quinnj: Yeah, we should probably support the `T` + `Union{T, Missing}` case pretty quickly.

rofinn: Is this something that could be resolved with `ChainedVector` type? This also impacts the `joinpartitions` solution as well.

rofinn: NOTE: This seems to be even more of an issue with the latest release with inline strings.

```julia
ArgumentError: string too large (36) to convert to InlineStrings.String31
Stacktrace:
[1] stringtoolong(T::Type, n::Int64)
@ InlineStrings ~/.julia/packages/InlineStrings/2HAHk/src/InlineStrings.jl:186
[2] InlineString
@ ~/.julia/packages/InlineStrings/2HAHk/src/InlineStrings.jl:209 [inlined]
[3] convert
@ ./strings/basic.jl:232 [inlined]
[4] convert
@ ./missing.jl:69 [inlined]
[5] getpoolidx
@ ~/.julia/packages/PooledArrays/jKDrZ/src/PooledArrays.jl:548 [inlined]
[6] push!(pv::PooledArrays.PooledVector{Union{Missing, InlineStrings.String31}, UInt32, Vector{UInt32}}, v::InlineStrings.String63)
@ PooledArrays ~/.julia/packages/PooledArrays/jKDrZ/src/PooledArrays.jl:594
[7] push!
@ ~/.julia/packages/SentinelArrays/VbnBp/src/chainedvector.jl:451 [inlined]
[8] append!(A::SentinelArrays.ChainedVector{Union{Missing, InlineStrings.String31}, PooledArrays.PooledVector{Union{Missing, InlineStrings.String31}, UInt32, Vector{UInt32}}}, B::PooledArrays.PooledVector{Union{Missing, InlineStrings.String63}, UInt32, Vector{UInt32}})
@ SentinelArrays ~/.julia/packages/SentinelArrays/VbnBp/src/chainedvector.jl:608
[9] (::TableOperations.var"#23#26"{Vector{SentinelArrays.ChainedVector}, DataFrames.DataFrameColumns{DataFrames.DataFrame}})(i::Int64)
@ TableOperations ~/.julia/packages/TableOperations/ErT2o/src/TableOperations.jl:312
[10] foreach(f::TableOperations.var"#23#26"{Vector{SentinelArrays.ChainedVector}, DataFrames.DataFrameColumns{DataFrames.DataFrame}}, itr::UnitRange{Int64})
@ Base ./abstractarray.jl:2141
[11] joinpartitions(x::Tables.Partitioner{Base.Generator{Vector{String}, Tables.var"#9#10"{S3DB.var"#220#221"}}})
```

PR #875, opened 2021-08-24
Improve time-to-first-read via precompile + despecialization

quinnj:
Been wanting/needing to do this for a while. Worked through the
excellent tutorials in
https://timholy.github.io/SnoopCompile.jl/stable/snoopi_deep_parcel/ to
help figure out cases where we were overspecializing certain methods
(`CSV.Context`) and generating useful precompile statements. On my
machine, this reduces the time-to-first-read (TTFR) from ~8s to ~1.5,
while package loading time goes from 3s to 10s, which both seem like
good tradeoffs. To avoid overspecialization of `CSV.Context` method,
which has a huge method body, we introduce a simple `Arg` wrapper struct
to kill inference and make the method signature concrete. The `@refargs`
macro conveniently wraps each arg in a _call_ with `Arg(x)`, and for a
_definition_, it annotates each arg as `x::Arg` and inserts an unwrap as
the first thing in the method body like `x = x[]::Int` where `x` was
declared as `x::Int` in the argument list.

One thing I noticed that I would still like to look into if time is that if we do `f = CSV.File(IOBuffer("a,b,c\n1,2,3\n\n"))` in a fresh session, it takes ~1.5s, but then if we call like `f = CSV.File(IOBuffer("a,b,c\n1,2,3\n\n"); header=false)`, it takes 3s to......recompile a bunch of stuff I guess? I'd love to figure out why that's causing so much recompiling.

### DIFF
--- a/src/chunks.jl
+++ b/src/chunks.jl
@@ -74,7 +74,7 @@ function Chunks(source;
     debug::Bool=false,
     parsingdebug::Bool=false)
 
-    ctx = Context(source, header, normalizenames, datarow, skipto, footerskip, transpose, comment, ignoreemptyrows, ignoreemptylines, select, drop, limit, buffer_in_memory, nothing, ntasks, tasks, rows_to_check, lines_to_check, missingstrings, missingstring, delim, ignorerepeated, quoted, quotechar, openquotechar, closequotechar, escapechar, dateformat, dateformats, decimal, truestrings, falsestrings, type, types, typemap, pool, downcast, lazystrings, stringtype, strict, silencewarnings, maxwarnings, debug, parsingdebug, false)
+    ctx = @refargs Context(source, header, normalizenames, datarow, skipto, footerskip, transpose, comment, ignoreemptyrows, ignoreemptylines, select, drop, limit, buffer_in_memory, nothing, ntasks, tasks, rows_to_check, lines_to_check, missingstrings, missingstring, delim, ignorerepeated, quoted, quotechar, openquotechar, closequotechar, escapechar, dateformat, dateformats, decimal, truestrings, falsestrings, type, types, typemap, pool, downcast, lazystrings, stringtype, strict, silencewarnings, maxwarnings, debug, parsingdebug, false)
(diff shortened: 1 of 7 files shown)


rikhuijzer:
> this reduces the time-to-first-read (TTFR) from ~8s to ~1.5,

Awesome! :heart:

timholy:
Nice work, @quinnj! With reference to https://discourse.julialang.org/t/csv-jl-fails-precompling-typeerror-in-type-expression-expected-unionall-got-type-parsers-options/67195, one thing I've noticed (but not yet updated the SnoopCompile docs for, sorry), is that to more easily support multiple Julia versions & architectures (e.g., 32-bit vs 64-bit), often a better strategy is to have your precompile file look like this:
```julia
function _precompile_()
    ccall(:jl_generating_output, Cint, ()) == 1 || return nothing
    x = my_pkg_object(args...)
    do_some_work(x, otherargs...)
end
```
for some tiny workloads typical of the data types you want your package to work on. This has the same effect as explicit `precompile` directives, except it's more thorough because any runtime-dispatches will also get precompiled without a separate `precompile` directive. And it better supports different Julia versions and architectures because it allows the compilation to proceed in the natural way on each machine.

In other words, rather than creating a workload for snooping via `@snoopi_deep` and then using `parcel` + `write`, better is to just paste your workload into the precompile file itself!

timholy:
It's the same as `precompile` because `precompile` is not a directive to "save my code in the cache": it simply forces the specialization to be compiled. Julia saves everything that's "owned" by the package in the cache automatically, whether it was compiled due to a `precompile` directive or simply by running code.

rikhuijzer:
Very nice @timholy. Now, I get what you meant when we were talking about it during JuliaCon. I did get it, but couldn't find any example anywhere. Looking forward to docs too :+1:

timholy:
Yeah, the problem with voluminous docs is that when you realize new stuff it's a bit daunting to update them. I really need to restructure it in more cookbook format ("do this...").

PR #876, opened 2021-08-24
Support SubArray{UInt8, 1, Vector{UInt8}} as valid source

quinnj:
This comes up sometimes when doing, for example,
`CSV.File(IOBuffer(strip(str)))`. Supporting `SubArray` where the parent
is `Vector{UInt8}` is very easy since we already support passing an
arbitrary `Vector{UInt8}` with custom starting/ending position.


src/utils.jl
```diff
@@ -149,10 +149,17 @@ consumeBOM(buf, pos) = (length(buf) >= 3 && buf[pos] == 0xef && buf[pos + 1] ==
...
-        return x.data, x.ptr, x.size, tfile
+    elseif x isa SubArray{UInt8, 1, Vector{UInt8}}
+        return parent(x), first(x.indices[1]), last(x.indices[1]), tfile
+    elseif x isa IOBuffer
+        if x.data isa Vector{UInt8}
+            return x.data, x.ptr, x.size, tfile
+        elseif x.data isa SubArray{UInt8}
+            x = x.data
+            return parent(x), first(x.indices), last(x.indices), tfile
+        end
```
vtjnash, line 162: This introduced a bug, since it is missing the `else` case here, and because all of these fields are private their implementation may change without warning. The typical API to call is `take!` to force the IOBuffer to destructively generate and return a `Vector{UInt8}` view of that current data

PR #897, opened 2021-09-10
Declare `isrowtable(::Rows) == true`

nickrobinson251:
- The existing code seems to be a typo, as
  as `rowtable` should return a `Vector{NamedTuple}`
  not a `Bool`. So i suspect we meant to
  define `isrowtable` (not `rowtable`).
- if `isrowtable`, then we get `rows(x) === x` for free

This is just something i happened to see when reading this part of the code for another reason.

### DIFF
--- a/src/rows.jl
+++ b/src/rows.jl
@@ -155,8 +155,7 @@ function Rows(source::ValidSources;
     )
 end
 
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Thanks!

PR #906, opened 2021-09-16
Allow passing function for types/pool keyword arguments

quinnj:
Implements #896. This makes the `types` and `pool` keyword arguments
more consistent with the `select` and `drop` keyword args in that they
accept functions of the form `(col_index, col_name) -> T`, where `types`
should return a type or `nothing`, which will result in the column type
being detected automatically. Similarly for `pool`, the function should
return a `Bool`, `Float64`, or `nothing`.


nickrobinson251:
Should the type constraints for `Rows` be loosened too? 
https://github.com/JuliaData/CSV.jl/blob/v0.9.3/src/rows.jl#L102
https://github.com/JuliaData/CSV.jl/blob/v0.9.3/src/rows.jl#L111

quinnj:
Good catch @nickrobinson251! Fixed in https://github.com/JuliaData/CSV.jl/commit/263e5c5f081fa73f4e67e2191b37cd208c901d42

PR #911, opened 2021-09-18
Add `validate` keyword for disabling argument checks

### LINKED ISSUE #910: Make validation on `types` parameter optional
#870 added validation to the `types` keyword, specifically requiring that all keys in the `Dict` also be columns in the CSV file.

It would be nice if such checking was optional, since it allows specifying types for _possible_ columns in a CSV. This is what we do in https://github.com/JuliaDynamics/Agents.jl/blob/master/src/submodules/io/csv_integration.jl, by analysing the field names of the agent struct to guess the columns and types in the CSV file. All the columns that are "guessed" may not appear in the CSV file.

Alternatively, is there a way to find the column names in the file from `CSV.Rows`? That would allow filtering the keys, and removing ones that aren't columns.

Ping https://github.com/JuliaDynamics/Agents.jl/pull/550

nickrobinson251:
Closes #910, follow-up to #870

- Not sure if we want to support this via yet another keyword, but if we do i think this is roughly how to do it
- Also not sure how this works with `Rows`... it seems right now there's only validation for `File` and not for `Rows`. Is that intended? and if so how's best to document this?
- And definitely not sure about the best name (`validate`? `checkargs`?  `checkcols`? ...?)

### DIFF
--- a/src/context.jl
+++ b/src/context.jl
@@ -104,9 +104,9 @@ end
 function checkinvalidcolumns(dict, argname, ncols, names)
     for (k, _) in dict
         if k isa Integer
-            (0 < k <= ncols) || throw(ArgumentError("invalid column number provided in `$argname` keyword argument: $k. Column number must be 0 < i <= $ncols as detected in the data"))
+            (0 < k <= ncols) || throw(ArgumentError("invalid column number provided in `$argname` keyword argument: $k. Column number must be 0 < i <= $ncols as detected in the data. To ignore invalid columns numbers in `$argname`, pass `validate=false`"))
         else
-            Symbol(k) in names || throw(ArgumentError("invalid column name provided in `$argname` keyword argument: $k. Valid column names detected in the data are: $names"))
+            Symbol(k) in names || throw(ArgumentError("invalid column name provided in `$argname` keyword argument: $k. Valid column names detected in the data are: $names. To ignore invalid columns names in `$argname`, pass `validate=false`"))
         end
     end
     return
@@ -140,6 +140,7 @@ struct Context
     maxwarnings::Int
     debug::Bool
     tempfile::Union{String, Nothing}
+    validate::Bool
     streaming::Bool
 end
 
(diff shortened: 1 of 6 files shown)


src/context.jl
```diff
@@ -247,6 +249,7 @@ end
     maxwarnings::Integer,
     debug::Bool,
     parsingdebug::Bool,
+    validate::Bool,
```
quinnj, line 252: I don't think we need to add this as a new field to `Context`, right? It should only ever be needed during the construction of the `Context` object

src/context.jl
```diff
@@ -442,7 +445,7 @@ end
                 customtypes = tupcat(customtypes, nonstandardtype(col.type))
             end
         end
-        checkinvalidcolumns(types, "types", ncols, names)
+        validate && checkinvalidcolumns(types, "types", ncols, names)
```
quinnj, line 448: I think it would be useful to mention that you can pass `validate=false` in the `checkinvalidcolumns` error just so people know what to do immediately if they hit the error.

nickrobinson251: yep, it's already added (up on L104), but let me know if it could be clearer; it currently reads:
```diff
-            (0 < k <= ncols) || throw(ArgumentError("invalid column number provided in `$argname` keyword argument: $k. Column number must be 0 < i <= $ncols as detected in the data"))
+            (0 < k <= ncols) || throw(ArgumentError("invalid column number provided in `$argname` keyword argument: $k. Column number must be 0 < i <= $ncols as detected in the data. To ignore invalid columns numbers in `$argname`, pass `validate=false`"))
         else
-            Symbol(k) in names || throw(ArgumentError("invalid column name provided in `$argname` keyword argument: $k. Valid column names detected in the data are: $names"))
+            Symbol(k) in names || throw(ArgumentError("invalid column name provided in `$argname` keyword argument: $k. Valid column names detected in the data are: $names. To ignore invalid columns names in `$argname`, pass `validate=false`"))
```

src/context.jl
nickrobinson251: that's a good point! And i see we do this for "`parsingdebug`" too (which is used in constructing the `Options`) -- i've pushed a commit removing it from the `Context` type itself.

quinnj (approved):
Thanks @nickrobinson251!

Issue #925, opened 2021-10-04
Method ambiguity in ` _promote` when both same `PooledVector` type


nickrobinson251:
```julia
julia> using CSV

julia> aa = IOBuffer("""
       a,b
       a,2
       a,3
       """)
IOBuffer(data=UInt8[...], readable=true, writable=false, seekable=true, append=false, size=12, maxsize=Inf, ptr=1, mark=-1)

julia> bb = IOBuffer("""
       a,b
       ab,22
       ab,33
       """)
IOBuffer(data=UInt8[...], readable=true, writable=false, seekable=true, append=false, size=16, maxsize=Inf, ptr=1, mark=-1)

julia> CSV.File([aa, bb])
ERROR: MethodError: _promote(::Type{PooledArrays.PooledVector{String3, UInt32, Vector{UInt32}}}, ::PooledArrays.PooledVector{String3, UInt32, Vector{UInt32}}) is ambiguous. Candidates:
  _promote(::Type{A}, x::A) where A in CSV at /Users/nick/.julia/packages/CSV/owrEo/src/utils.jl:171
  _promote(::Type{PooledArrays.PooledVector{T, R, RA}}, x) where {T, R, RA} in CSV at /Users/nick/.julia/packages/CSV/owrEo/src/utils.jl:182
Possible fix, define
  _promote(::Type{PooledArrays.PooledVector{T, R, RA}}, ::PooledArrays.PooledVector{T, R, RA}) where {T, R, RA}
Stacktrace:
 [1] chaincolumns!(a::Any, b::Any)
   @ CSV ~/.julia/packages/CSV/owrEo/src/utils.jl:206
 [2] CSV.File(sources::Vector{IOBuffer}; source::Nothing, kw::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ CSV ~/.julia/packages/CSV/owrEo/src/file.jl:1010
 [3] CSV.File(sources::Vector{IOBuffer})
   @ CSV ~/.julia/packages/CSV/owrEo/src/file.jl:985
 [4] top-level scope
   @ REPL[19]:1

(jl_weYOTi) pkg> st
      Status `/private/var/folders/hx/1h0bbkfd18d4n1qrnwmrl4j00000gn/T/jl_weYOTi/Project.toml`
  [336ed68f] CSV v0.9.6
```

nickrobinson251:
little reproducer
```julia
[...]

PR #926, opened 2021-10-04
Fix ambiguity of "promoting" a PooledVector to its same type

nickrobinson251:
- fix issue #925


nickrobinson251:
oof, ambiguities

PR #949, opened 2021-12-14
Avoid promoting inline string types larger than String31

### LINKED ISSUE #945: Possible string type detection bug
@quinnj it seems that in some cases `CSV.read` creates `String63` columns under default settings (and AFAICT it should not - `String` should be used in such a case as we are using InlineStrings.jl up to `String31` by default).

Here is a notebook https://github.com/JuliaAcademy/DataFrames/blob/main/3.%20Working%20with%20text%20files.ipynb reproducing the issue. See cell [12] for the result and cell [10] for the command I use to read in the data. To get the data use the code in cell [5].

quinnj:
Fixes #945. The core issue here is an inconsistency between the initial
type detection code and the later-on type promotion code while parsing.
The type detection code had a limit setup so that inline string column
types weren't allowed larger than `String31`. The promotion code,
however, allowed inline string types to continue to promote up to
`String255`. That means if a column type was at least initiall detected
as some smaller-than-`String31` type, then later on a value larger than
31 bytes was parsed, it would continue to promote up to larger inline
string types. The change proposed in this PR is to limit the promotion
code to be more consistent with the type detection code: if a parsed
value "overflows" the `String31` type, we'll just promote to a regular
`String` instead of promoting to larger inline string types.

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -869,9 +869,9 @@ function parsevalue!(::Type{type}, buf, pos, len, row, rowoffset, i, col, ctx)::
                         code = trytopromote!(type, newT, buf, pos, len, col, row)
                         newT = _widen(newT)
                     end
-                elseif type === InlineString1 || type === InlineString3 || type === InlineString7 || type === InlineString15 || type === InlineString31 || type === InlineString63 || type === InlineString127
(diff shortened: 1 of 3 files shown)


bkamins:
Thank you! Are you planning to make a patch release or you are waiting for some additional changes before this?

Issue #951, opened 2021-12-18
Option to strip whitespace from strings


tp2750:
It would be good to have an option to strip leading and trailing whitespace from text columns.

This us useful if the input table has been padded to make the delimiters line up (eg orgmode tables). 

Currently we have:

```{julia}
data = """
| Name |  Zip |
| Joe  |  123 |
| Anne | 1234 |
"""
CSV.File(IOBuffer(data); delim='|', normalizenames=true)
2-element CSV.File:
 CSV.Row: (Column1 = missing, Name = " Joe  ", Zip = 123, Column4 = missing)
 CSV.Row: (Column1 = missing, Name = " Anne ", Zip = 1234, Column4 = missing)
```

It would be great to be able to add a keyword option like "stripwitespace=true" and get

```{julia}
data = """
| Name |  Zip |
| Joe  |  123 |
| Anne | 1234 |
"""
CSV.File(IOBuffer(data); delim='|', normalizenames=true, stripwhitespace=true)
2-element CSV.File:
 CSV.Row: (Column1 = missing, Name = "Joe", Zip = 123, Column4 = missing)
 CSV.Row: (Column1 = missing, Name = "Anne", Zip = 1234, Column4 = missing)
```

In R this is similar to `strip.white=TRUE` in `read.table()`

Conceptually, this would be applying `strip()` to all string elements.

I'll be happy to try to implement this, if it is found useful, and I can get a bit of guidance.

quinnj:
Yeah, this shouldn't be too hard to support. The change would come in Parsers.jl. We could add a new "option" to the `Parsers.Options` struct to strip whitespace for strings (we already strip whitespace by default for non-strings). Then the implementation would happen in [this file](https://github.com/JuliaData/Parsers.jl/blob/main/src/strings.jl#L3). You'll notice there's already some code around stripping leading whitespace and that's because if we have a string like `  " abc " `, the leading/trailing whitespace before/after the double quotes will be stripped, while the "inner" whitespace (the single spaces before/after `abc`) will be preserved. So we would just want to check if `options.stripwhitespace` and then reset `vstartpos` to `pos` after we've stripped whitespace.

Thanks for the report and offer to help!

tp2750:
Thanks for the hints @quinnj . I'll give it a shot.

tp2750:
I think I've fixed it :smile: 

At least the tests of CSV are all green after I added these 2:

```
data = """
| Name       |  Zip |
[...]

PR #953, opened 2021-12-20
951 whitespace

tp2750:
Add option to strip whitespace from string fields, as discussed in JuliaData/CSV.jl#951


test/basics.jl
```diff
@@ -748,4 +748,15 @@ f = CSV.File(IOBuffer("a,b\n1,2\n3,"))
...
 
+# 951
+data = """
+| Name       |  Zip |
+| Joe        |  123 |
+| Mary Anne  | 1234 |
+"""
+f = CSV.File(IOBuffer(data); delim='|', normalizenames=true, stripwhitespace=false)
+@test f.Name[1] == " Joe        "
+f = CSV.File(IOBuffer(data); delim='|', normalizenames=true, stripwhitespace=true)
```
quinnj, line 759: You can remove `normalizenames` in this case, right? Since `stripwhitespace` will result in column names like `Name` and `Zip`, right? It might be good to clarify this in the keyword doc; i.e. that the whitespace stripping also applies to parsing the column names.

tp2750: Good point. I'll update the test and documentation.

src/keyworddocs.jl
```diff
@@ -29,6 +29,7 @@ const KEYWORD_DOCS = """
   * `dateformat::Union{String, Dates.DateFormat, Nothing, AbstractDict}`: a date format string to indicate how Date/DateTime columns are formatted for the entire file; if given as an `AbstractDict`, date format strings to indicate how the Date/DateTime columns corresponding to the keys are formatted. The Dict can map column index `Int`, or name `Symbol` or `String` to the format string for that column.
   * `decimal='.'`: a `Char` indicating how decimals are separated in floats, i.e. `3.14` uses `'.'`, or `3,14` uses a comma `','`
   * `truestrings`, `falsestrings`: `Vector{String}`s that indicate how `true` or `false` values are represented; by default `"true", "True", "TRUE", "T", "1"` are used to detect `true` and `"false", "False", "FALSE", "F", "0"` are used to detect `false`; note that columns with only `1` and `0` values will default to `Int64` column type unless explicitly requested to be `Bool` via `types` keyword argument
+  * `stripwhitespace=false`: if true, leading and trailing whitespace is stripped from string fields
```
quinnj, line 32:
```suggestion
  * `stripwhitespace=false`: if true, leading and trailing whitespace are stripped from string values, including column names
```

test/basics.jl
quinnj:
```suggestion
f = CSV.File(IOBuffer(data); delim='|', stripwhitespace=true)
```

quinnj (approved):
Thanks @tp2750 for doing this!

tp2750:
Thanks a lot @quinnj .
Good call not to wait for me doing the small edits of the documentation and test :smile: 
This has been a good experience.
Thanks for all your work on CSV.jl (and other Julia packages), for getting this feature in, and also for helping me get at a glimpse into the internals of CSV.jl.

JackDunnNZ:
While upgrading to CSV 0.10, I ran into the following error (with Parsers 2.1.3). Looks like this PR should have bumped compat for Parsers to 2.2 to make sure the new flag is available:

```julia
ERROR: LoadError: MethodError: no method matching Parsers.Options(::Missing, ::UInt8, ::UInt8, ::UInt8, ::UInt8, ::UInt8, ::UInt8, ::UInt8, ::Vector{String}, ::Vector{String}, ::Nothing, ::Bool, ::Bool, ::Nothing, ::Bool, ::Bool, ::Bool)
```

quinnj:
Oops, thanks for reporting @JackDunnNZ! Should be fixed here: https://github.com/JuliaData/CSV.jl/commit/1061e24a9e504b323a9c4e77ad8b717ddd2105ce

PR #955, opened 2021-12-21
Improve performance of reading files with duplicate column names

wentasah:
I need to load a file with 30k columns, 10k of these have the same name. Currently, this is practically impossible because `makeunique()`, which produces unique column names, has cubic complexity.

This changes the algorithm to use Set and Dict to quickly look up the existence of columns and to cache the last numeric suffix used to uniquify column names.

Care has been taken to ensure that columns are named the same way as before. To that extent, additional tests were added in the first commit. `makeunique` was updated in the second commit to make it easy to verify that the tests pass both before and after the change.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -349,17 +349,20 @@ function makeunique(names)
     set = Set(names)
     length(set) == length(names) && return Symbol[Symbol(x) for x in names]
     nms = Symbol[]
+    nextsuffix = Dict{eltype(names), UInt}()
     for nm in names
-        if nm in nms
-            k = 1
+        if haskey(nextsuffix, nm)
+            k = nextsuffix[nm]
             newnm = Symbol("$(nm)_$k")
-            while newnm in set || newnm in nms
+            while newnm in set || haskey(nextsuffix, newnm)
                 k += 1
                 newnm = Symbol("$(nm)_$k")
             end
+            nextsuffix[nm] = k + 1
             nm = newnm
         end
         push!(nms, nm)
+        nextsuffix[nm] = 1
     end
     @assert length(names) == length(nms)
     return nms
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -748,4 +748,14 @@ f = CSV.File(IOBuffer("a,b\n1,2\n3,"))
 @test f.a == [1, 3]
 @test isequal(f.b, [2, missing])
 
+# duplicate column names
+f = CSV.File(IOBuffer("a,a,a\n"))
+@test f.names == [:a, :a_1, :a_2]
+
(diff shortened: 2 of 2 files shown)


src/utils.jl
```diff
@@ -349,17 +349,21 @@ function makeunique(names)
     set = Set(names)
     length(set) == length(names) && return Symbol[Symbol(x) for x in names]
     nms = Symbol[]
+    nmsset = Set{eltype(names)}()
```
quinnj, line 352: It seems to me, without having actually tried to code it out, that we could avoid needing this separate `nmsset` and just re-use the `set` from above? And then we would check `if nm in set` below, then just `while newnm in set`, then finally do `push!(set, nm)`. That way we avoid having two `Set`s that basically do the same thing?

I remember this code being a little subtle in a few cases though, so perhaps I'm forgetting an oddity that prevents us doing it that way.

wentasah: I was also thinking about that, but that would change how the columns are named in some cases. And this could break existing code relying on the uniquified names.

Consider the following code:
```julia
CSV.makeunique([:a :a])
```
Without `nmsset`, as you propose, the result is:
```
2-element Vector{Symbol}:
 :a_1
 :a_2
```
while with `nmsset` it is:
```
2-element Vector{Symbol}:
 :a
 :a_1
```

wentasah: It seems it will be possible to use `suffixes` Dict instead of `nmsset`. The result will be slightly shorter code. I'm currently testing it and push it soon.

src/utils.jl
```diff
@@ -349,17 +349,20 @@ function makeunique(names)
     set = Set(names)
     length(set) == length(names) && return Symbol[Symbol(x) for x in names]
     nms = Symbol[]
+    nextsuffix = Dict{eltype(names), UInt}()
     for nm in names
-        if nm in nms
-            k = 1
+        if nm in keys(nextsuffix)
```
quinnj, line 354:
```suggestion
        if haskey(nextsuffix, nm)
```
using `haskey` is more idiomatic

src/utils.jl
```diff
@@ -349,17 +349,20 @@ function makeunique(names)
...
+            k = nextsuffix[nm]
             newnm = Symbol("$(nm)_$k")
-            while newnm in set || newnm in nms
+            while newnm in set || newnm in keys(nextsuffix)
```
quinnj, line 357:
```suggestion
            while newnm in set || haskey(nextsuffix, newnm)
```

wentasah: Thanks for suggestion. Updated.

quinnj:
Thanks @wentasah!

PR #961, opened 2022-01-14
Correct error writing struct

jcunwin:
When using `CSV.write` on a array of `struct`s, I want to override the column names of the struct in the CSV header.
```jl
using Dates
using CSV

struct StructType
    adate::Date
    astring::Union{String, Nothing}
    anumber::Union{Real, Nothing}
end

structtable = [StructType(Date("2021-12-01"), "string 1", 123.45), StructType(Date("2021-12-31"), "string 2", 456.78)]

io = IOBuffer()

CSV.write(io, structtable; header=["Date Column", "String Column", "Number Column"])
```
An exception occurs because `write.jl` tries to use the specified header value (`"Date Column"`) to retrieve the struct's property value instead of the struct property name (`:adate`), resulting in an exception. E.g.:

```
CSV.write: Error During Test at /home/john/Code/investments/CSV.jl/test/write.jl:41
  Got exception outside of a @test
  type StructType has no field Date Column
  Stacktrace:
    [1] getproperty
      @ ./Base.jl:42 [inlined]
    [2] getcolumn(x::StructType, nm::Symbol)
      @ Tables ~/.julia/packages/Tables/M26tI/src/Tables.jl:102
    [3] getcolumn
      @ ~/.julia/packages/Tables/M26tI/src/tofromdatavalues.jl:74 [inlined]
    [4] eachcolumn
      @ ~/.julia/packages/Tables/M26tI/src/utils.jl:86 [inlined]
etc...
```
I.e. It tries something like:
```jl
getproperty(Symbol("Date Column"))
```
instead of something like:
```jl
getproperty(Symbol(:adate))
```
A fix is possible in `write.jl` in the `write` function. I.e. instead of using the header names to derive the schema:

```jl
# handle unknown schema case
function write(::Nothing, rows, file, opts;
        append::Bool=false,
        compress::Bool=false,
        header::Union{Bool, Vector}=String[],
        bufsize::Int=2^22
    )

#snip

    names = header isa Bool || isempty(header) ? propertynames(row) : header
    sch = Tables.Schema(names, nothing)    # <- Problem is here - wrongly using header values
    cols = length(names)

#snip

    return file
end
```

... use the data schema directly:

```jl
# handle unknown schema case
function write(::Nothing, rows, file, opts;
        append::Bool=false,
        compress::Bool=false,
        header::Union{Bool, Vector}=String[],
        bufsize::Int=2^22
    )

#snip

    names = header isa Bool || isempty(header) ? propertynames(row) : header
    sch = Tables.Schema(propertynames(row), nothing)   # <- fix is here - use property names
    cols = length(names)

#snip

    return file
end
```
I have created a fork with a code fix and a test case here:

https://github.com/jcunwin/CSV.jl


src/write.jl
```diff
@@ -262,7 +262,7 @@ function write(::Nothing, rows, file, opts;
     end
     row, st = state
     names = header isa Bool || isempty(header) ? propertynames(row) : header
-    sch = Tables.Schema(names, nothing)
+    sch = Tables.Schema(propertynames(row), nothing)
```
quinnj, line 265:
```suggestion
    sch = Tables.Schema(Tables.columnnames(row), nothing)
```

test/write.jl
```diff
@@ -17,6 +17,14 @@ const table_types = (
     col8=weakrefs,
 )
 
+struct StructType
+    adate::Date
+    astring::Union{String, Nothing}
+    aumber::Union{Real, Nothing}
+end
+
+#structtable = [StructType(Date("2021-12-01"), "string 1", 123.45), StructType(Date("2021-12-02"), "string 2", 456.78)]
```
quinnj, line 26:
```suggestion
```

quinnj:
Thanks @jcunwin for this!

PR #964, opened 2022-01-17
Fix use of limit in multithreaded parsing

### LINKED ISSUE #963: Problem with SentinelArrays.ChainedVector when limit/skipto is set
@quinnj:
Start Julia with multiple threads.

Run:
```
using DataFrames
using CSV
df = DataFrame(x = repeat(["a", "b"], 10^6), y = rand(2*10^6));
f1 = tempname();
CSV.write(f1, df);
df2 = CSV.read(f1, DataFrame, header = false, skipto = 1001, limit = 10000);
sort(df2, 1)
```

The problem is that in both columns `SentinelArrays.ChainedVector` the length of `arrays` does not match `inds`. Here is an example for `Column2`:

```
julia> df2.Column2.inds
4-element Vector{Int64}:
  2853
  5706
  8559
 10000

julia> cumsum(length.(df2.Column2.arrays))
4-element Vector{Int64}:
 2853
 5706
 8559
 9970
```
and you see that there are 30 elements missing.

If you reimplemented the `getindex` like:
```
function Base.getindex(A::ChainedVector, i::Integer)
    chunk, ix = index(A, i)
    x = A.arrays[chunk][ix]
    return x
end
```
to enable bounds checking you get bounds error when trying to work with `df2`.

quinnj:
Fixes #963. The issue here is that although we were adjusting the # of
rows to a provided limit when multithreaded parsing, we failed to adjust
the actual column arrays to the correct size. This was an issue when we
converted from the old `CSV.Column` custom array type to returning
"normal" arrays in the 0.7 -> 0.8 transition. With `CSV.Column`, we just
passed the final row total and it adjusted the size dynamically, without
physically resizing the underlying array. With regular arrays, however,
we need to ensure the array gets resized appropriately. This became more
apparent in the recent pooling change that was released since it
actually became a silenced BoundsError because of the use of `@inbounds`
in the new `checkpooled!` routine. I've taken out those `@inbounds` uses
for now to be more conservative. The fix is fairly straightforward in
that if we adjust our final row down to a user-provided limit, then we
loop over the parsing tasks and "accumulate" rows until we hit the limit
and then resize or `empty!` columns as appropriate.

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -243,6 +243,23 @@ function File(ctx::Context, @nospecialize(chunking::Bool=false))
         finalrows = sum(rows)
         if ctx.limit < finalrows
             finalrows = ctx.limit
+            # adjust columns according to limit
+            acc = 0
+            for i = 1:ntasks
+                if acc + rows[i] > finalrows
+                    # need to resize this tasks columns down
+                    if finalrows - acc > 0
+                        for col in pertaskcolumns[i]
+                            resize!(col.column, finalrows - acc)
+                        end
+                    else
+                        for col in pertaskcolumns[i]
+                            empty!(col.column)
+                        end
+                    end
+                end
+                acc += rows[i]
+            end
         end
         # ok, all the parsing tasks have finished and we've promoted their types w/ the top-level columns
         # so now we just need to finish processing each column by making ChainedVectors of the individual columns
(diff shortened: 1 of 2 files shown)


bkamins:
you probably need to rebase this PR :)

Issue #974, opened 2022-01-30
First call of CSV.File very slow


ufechner7:
I have the following code:
```julia
@time using CSV

const input="""
time,ping
1,25.7
2,31.8
"""

function read_csv(in)
    io = IOBuffer(in)
    @time file = CSV.File(io)
    close(io)
    file
end

file = read_csv(input)
```
Which give the following output:
```julia
julia> @time include("bench4.jl")
  2.255220 seconds (5.69 M allocations: 328.290 MiB, 6.23% gc time, 91.37% compilation time)
 11.489286 seconds (2.32 M allocations: 96.562 MiB, 99.98% compilation time)
 17.222498 seconds (53.52 M allocations: 2.279 GiB, 4.79% gc time, 98.84% compilation time)
2-element CSV.File:
 CSV.Row: (time = 1, ping = 25.7)
 CSV.Row: (time = 2, ping = 31.8)
```
```julia
julia> versioninfo()
Julia Version 1.7.1
Commit ac5cc99908 (2021-12-22 19:35 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
```
So the first call of this toy example needs 11.5 s on a pretty fast machine.

I consider this a bug, it should be possible to bring this time down by a factor of 2 or 3 just by using
proper pre-compilation. Am I missing something?

rafaqz:
Type instability pretty much everywhere in very large function. There are easy improvements everywhere just by adding function barriers and splitting things up. 

For example, moving the `ctx.threaded` code into another function from file.jl line 229 along shave the time from 16 to 12 second on my laptop. There is no need to compile that code in single threaded contexts.

ufechner7:
The following code is about 10 times as fast for this toy example:
```
@time using DataFrames, DelimitedFiles

const input="""
time,ping
1,25.7
2,31.8
"""

function read_csv(inp)
    @time data, header = readdlm(IOBuffer(inp), ',',header=true)
    # @time df = DataFrame(data, vec(header))
    @time df = identity.(DataFrame(ifelse.(data .== "", missing, data), vec(header)))
    @time df[!,:time] = convert.(Int64,df[:,:time])
    df
end

df = read_csv(input)
```
[...]

PR #975, opened 2022-01-31
improve time to first csv load

rafaqz:
This PR aims to at least partially solve #974

This first commit demonstrates that simply breaking up large unstable functions can greatly improve compilation time. Splitting `File` here into threaded and single threaded contexts means only the one we are using gets compiled. For me this shaves the time from #974 from 16seconds to 12 seconds. Without changing anything else.

There are probably a lot more potential for this. The functions in file.jl all seem to be too long.


bkamins:
bravo!

rafaqz:
Fixing type stability and adding precompilination runs brings the total load time down to 7 seconds, and `using CSV` is 1 second.

There is another second available by removing the `@refargs` macro so `Context` precompiles, but I'm not sure what the purpose of this macro is?

quinnj:
Sorry for not being able to review earlier; I think this is exciting stuff! One comment I'll make is that a lot of the current code organization was done in an effort to avoid _recompilation_ costs as much as possible. It's become clear to me now that doing that has brought on a pretty severe cost for TTFR, but I'd also like to keep that in mind as we make efforts to reduce TTFR. I.e. some of the changes here mean that if you slightly change a keyword argument, we have to pay significant compilation costs again, which hopefully we can avoid. In CSV.jl early days, we were at the opposite end of the spectrum where every single call to `CSV.File` almost guaranteed recompiling everything; but I think I agree that we've probably swung back too far the other way by compiling _too_ much when you initially call `CSV.File` and we can allow more incremental compilation as various features are actually used/needed while parsing.

As a reviewer comment, it would be nice to probably split this up into separate PRs? As-is, it's very hard to review since there's so much moving/changing around.

rafaqz:
I figured something like that might have been the case. If we can get precompilation working nearly everywhere and functions are broken up enough that they are decoupled, we should hopefully only see minor recompilation for small changes. Its really the file.jl internals that take a long time and they are precompiling for all common types now it seems. `Context` takes under a second the first time, so subsequent changes shouldn't be too bad either. 

I know this PR is currently pretty messy and buggy... its pretty much a blind search for type stability with Cthulhu.jl. 

Unfortunately its hard to split up the PR while working on it because precompilation only works past some certain threshold of stability and function size (that I don't totally understand to be honest), and before that most things have no noticeable effect. Once I can lock down everything required to make this work I can go back and split it all up into separate changes.

rafaqz:
Precompile in Parsers.jl helps a lot here: https://github.com/JuliaData/Parsers.jl/pull/108

TTFX with both of these changes is now about 4 seconds.

rafaqz:
There another 30-40% TTFX improvement in some of this code, but some changes are unnecessary now - this branch had larger gains without the Paraers.jl fix.

The changes here that still make sense will be more controversial for more modest improvements, so won't be high priority for me.

rafaqz:
Probably restructuring Context and the preloaded of that would be the most obvious next step.

ViralBShah:
I suppose this is outdated and can be closed.

rafaqz:
I didnt mention here but after the later fix to Parsers.jl (removing just 4 `@inline`s 😂) this PR stopped doing much:

https://github.com/JuliaData/Parsers.jl/pull/1140

All the `@noinlines` and function barriers were effective because of the constants that Parsers.jl inlined. There are probably a few type stability things here worth keeping, but may starting a new PR and benchmarking without the Parsers.jl problems would make more sense.

PR #977, opened 2022-02-04
remove init and precompile

rafaqz:
This is a much simpler PR than the last one, and gives quite a good TTFX improvment for no work.

The `__init__()` method here barely helps with `File` or `Context`, but costs a few seconds with `using`.

Perversely, on 1.7 `_precompile_()` actually makes the first load slower.

Removing them both gives a couple of seconds improvement in time to `using CSV; File(filename)`


rafaqz:
Hmm only seems to work when other changes are made already. So I will do this later.

Issue #976, opened 2022-01-31
Cannot load CSV.jl at all (precompilation)


mileslucas:
I can't use (seemingly) any version of CSV.jl on my current set up. I am using Mac OS X Monterey with a Julia version built for x86

```julia
julia> versioninfo()
Julia Version 1.7.1
Commit ac5cc99908 (2021-12-22 19:35 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Apple M1 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, westmere)
Environment:
  JULIA_NUM_THREADS = auto
```

Whenever I 
```julia
using CSV
```
I get stuck precompiling and it never ends. Like, I ran it overnight without success. `ctrl-c` gives "Interrupted: Exiting precompilation...", so no useful information there.

I've tried reverting to `0.9` versions, no success, I've tried explicitly trying `0.10.0` `0.10.1` and `0.10.2` and none of them work. I've tried checking out from `main` without success.

ufechner7:
What is the output of the command:
```
]st
```
?

mileslucas:
It doesn't work in a clean environment-

```julia
(@v1.7) pkg> activate --temp
  Activating new project at `/var/folders/q6/j_0w4_mx13jg_7mw9__902f40000gn/T/jl_P39Qal`

(jl_P39Qal) pkg> add CSV
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
    Updating `/private/var/folders/q6/j_0w4_mx13jg_7mw9__902f40000gn/T/jl_P39Qal/Project.toml`
  [336ed68f] + CSV v0.10.2
    Updating `/private/var/folders/q6/j_0w4_mx13jg_7mw9__902f40000gn/T/jl_P39Qal/Manifest.toml`
  [336ed68f] + CSV v0.10.2
  [944b1d66] + CodecZlib v0.7.0
  [34da2185] + Compat v3.41.0
  [9a962f9c] + DataAPI v1.9.0
  [e2d170a0] + DataValueInterfaces v1.0.0
  [48062228] + FilePathsBase v0.9.17
  [842dd82b] + InlineStrings v1.1.2
  [82899510] + IteratorInterfaceExtensions v1.0.0
  [69de0a69] + Parsers v2.2.1
  [2dfb63ee] + PooledArrays v1.4.0
[...]

PR #979, opened 2022-02-14
fix  Parsers version to unbreak precompilation

oscardssmith:
fixes https://github.com/JuliaData/CSV.jl/issues/976


quinnj:
I dont' suspect this PR is the culprit, but since merging, I'm seeing consistent failures on `main` branch for windows-latest platform. Planning to try and investigate what's going on before tagging a release.

PR #989, opened 2022-03-08
fix ccall usage on CSV 0.5

KristofferC:
There are quite a few packages that are stuck on CSV 0.5 and it starts to error
on Julia 1.8 due to the faulty `ccall` usage here.

It would be good if a `release-0.5` branch could be created, this merged into it
and a new 0.5 release tagged from it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CSV"
 uuid = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 authors = ["Jacob Quinn <quinn.jacobd@gmail.com>"]
-version = "0.5.26"
+version = "0.5.27"
 
 [deps]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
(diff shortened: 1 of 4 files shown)


KristofferC:
cc @quinnj

quinnj:
Yikes; why are people stuck on 0.5? That seems so ancient at this point.

KristofferC:
It is these packages:

```
 "StaticTrafficAssignment"
 "MarketCycles"
 "EconJobMarket"
 "FeatureSelectors"
 "SocialSolver"
 "BIDSTools"
 "GeoStatsDevTools"
 "ExSup"
 "DeIdentification"
 "MambaModels"
 "PerfChecker"
 "PredictMDExtra"
 "SOM"
 "ClimateTools"
 "ClimatePlots"
 "Qwind"
 "PowerGraphics"
 "AnyMOD"
 "AtomicGraphNets"
```

Some seems to be on 0.6 as well.

> Yikes; why are people stuck on 0.5? 

Most people don't really follow up on when their dependencies break their API so any such breakage leaves quite a few package more or less permanently stuck on old versions.

It is of course possible to just ignore them, perhaps they aren't used very much.

PR #1004, opened 2022-05-16
Show an example on top of `CSV.read`

rikhuijzer:
I was looking through the documentation of this package and noticed that there is no example for `CSV.read`. Instead it just lists all the arguments. I suggest to add a small example which answers:

1. "How do I open a `CSV` file as a `DataFrame`?" since the majority of readers will be looking for this
2. "How do I pass one of these listed arguments to the `CSV.read` function?"

Maybe we can move the `CSV.read` also to the top of the page at <https://csv.juliadata.org/stable/reading.html>? That would make the information more consistent with the <https://csv.juliadata.org/stable/writing.html>. The writing page is excellent by the way. Very clear.

### DIFF
--- a/src/CSV.jl
+++ b/src/CSV.jl
@@ -82,6 +82,30 @@ include("write.jl")
 Read and parses a delimited file or files, materializing directly using the `sink` function. Allows avoiding excessive copies
 of columns for certain sinks like `DataFrame`.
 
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Thanks @rikhuijzer!

PR #1006, opened 2022-06-02
Add links to alternatives to the readme

mcabbott:
As discussed here https://discourse.julialang.org/t/how-do-i-know-if-a-package-is-good/82133, it might be nice if this package linked to alternative ways to read CSV files. Really all packages should do this, but this one is what you find if you google "csv julia"... and I bet that many people googling that just want DelimitedFiles. 

I'm not really qualified to write the list of fancier alternatives, since DelimitedFiles does what I need right now. But @juliohm @sl-solution @chriselrod from discourse thread may have better ideas. And in particular about what most important differences should be mentioned.

I'm especially unqualified to list Python or R alternatives. That's less obviously necessary, but their names are terms people might search for. 

Xref https://github.com/JuliaPy/Pandas.jl/issues/87 about Pandas.jl -> DataFrames.jl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,9 @@ Contributions are very welcome, as are feature requests and suggestions. Please 
 
 There are several other packages for reading CSV files in Julia, which may suit your needs better:
 
-* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/),
-  which is perfect for quickly reading small files.
+* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/).
+  This returns a `Matrix` rather than a [Tables.jl](https://github.com/JuliaData/Tables.jl)-style container, thus works best for files of homogenous element type. 
+  On large files, CSV.jl will be much faster.
 
 * [CSVFiles.jl](https://github.com/queryverse/CSVFiles.jl) uses the [FileIO.jl](https://github.com/JuliaIO/FileIO.jl) API 
   into any [IterableTables.jl](https://github.com/queryverse/IterableTables.jl) sink.
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Contributions are very welcome, as are feature requests and suggestions. Please 
 
 There are several other packages for reading CSV files in Julia, which may suit your needs better:
 
-* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/).
+* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/), at least until Julia 1.8.
   This returns a `Matrix` rather than a [Tables.jl](https://github.com/JuliaData/Tables.jl)-style container, thus works best for files of homogenous element type. 
   On large files, CSV.jl will be much faster.
 
--- a/README.md
+++ b/README.md
@@ -40,3 +40,16 @@ Contributions are very welcome, as are feature requests and suggestions. Please 
 [codecov-url]: https://codecov.io/gh/JuliaData/CSV.jl
 
 [issues-url]: https://github.com/JuliaData/CSV.jl/issues
+
+## Alternatives
+
+There are several other packages for reading CSV files in Julia, which may suit your needs better:
+
+* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/),
+  which is perfect for quickly reading small files.
+
+* [DLMReader.jl](https://github.com/sl-solution/DLMReader.jl) also aims to be fast for large files. 
+  Closely associated with [InMemoryDatasets.jl](https://github.com/sl-solution/InMemoryDatasets.jl).
+
+Python and R also have CSV readers in the standard library. Fast packages include...
+
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ There are several other packages for reading CSV files in Julia, which may suit 
   into any [IterableTables.jl](https://github.com/queryverse/IterableTables.jl) sink.
   The package uses [TextParse.jl](https://github.com/queryverse/TextParse.jl) for parsing.
 
-* [DLMReader.jl](https://github.com/sl-solution/DLMReader.jl) also aims to be fast for large files. 
-  Closely associated with [InMemoryDatasets.jl](https://github.com/sl-solution/InMemoryDatasets.jl) rather than [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl)
+* [DLMReader.jl](https://github.com/sl-solution/DLMReader.jl) also aims to be fast for large files,
+  closely associated with [InMemoryDatasets.jl](https://github.com/sl-solution/InMemoryDatasets.jl).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,3 +40,19 @@ Contributions are very welcome, as are feature requests and suggestions. Please
 [codecov-url]: https://codecov.io/gh/JuliaData/CSV.jl
 
 [issues-url]: https://github.com/JuliaData/CSV.jl/issues
+
+## Alternatives
+
+There are several other packages for reading CSV files in Julia, which may suit your needs better:
+
+* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/),
+  which is perfect for quickly reading small files.
```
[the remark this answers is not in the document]
mcabbott: Good point. Maybe something like this:
```suggestion
* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/).
  This returns a `Matrix` rather than a `DataFrame`, thus works best for files of homogenous element type. 
  On large files, CSV.jl will be much faster.
```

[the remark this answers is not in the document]
nickrobinson251: A couple more notes:

CSV doesn't return a DataFrame by default, it returns a CSV.File, so we should be careful not to imply it does

DelimitedFiles won't be a stdlib in upcoming Julia versions, I think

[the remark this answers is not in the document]
chriselrod: Looks good.

README.md
```diff
@@ -40,3 +40,19 @@ Contributions are very welcome, as are feature requests and suggestions. Please
...
+
+* [CSVFiles.jl](https://github.com/queryverse/CSVFiles.jl) uses the [FileIO.jl](https://github.com/JuliaIO/FileIO.jl) API 
+  into any [IterableTables.jl](https://github.com/queryverse/IterableTables.jl) sink.
+  The package uses [TextParse.jl](https://github.com/queryverse/TextParse.jl) for parsing.
```
mcabbott, line 53: I copied this text from https://github.com/queryverse/TextParse.jl . But it's not terribly helpful for knowing which to choose. Both packages use Tables.jl, what's the most obvious sign you ought to use one or the other?

README.md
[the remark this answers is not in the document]
brenhinkeller: > DelimitedFiles won't be a stdlib in upcoming Julia versions 
That's slightly horrifying

README.md
mcabbott: Paging @davidanthoff -- how best should CSVFiles.jl be described?

quinnj:
Yeah..........this is mostly fine. I'm hopeful that by the CSV.jl 1.0 release, we can get the time-to-first-read to be really competitive with DelimitedFiles.jl and then I don't think there's really any reason to use it.

DLMReader.jl is a great reference because it supports some more exotic parsing configurations if you need to get really custom.

I'd prefer we remove the reference to CSVFiles.jl/TextParse.jl; they haven't been updated or had any real work done for a long time, and they don't provide any kind of functionality not supported in CSV.jl.

mcabbott:
> really competitive with DelimitedFiles.jl and then I don't think there's really any reason to use it

Unless you want a matrix not a dataframe, right? There is a place for loading/saving 10 numbers without figuring out any complicated types.

> prefer we remove the reference to CSVFiles.jl/TextParse.jl; they haven't been updated

I guess my vote is to then say roughly that. It's useful information if you are trying to figure out how all these packages relate to each other. Otherwise you have to try to infer from the dates... is X not mentioned by Y because it's the newer nice thing which didn't exist when Y's summary was written?

quinnj:
Yeah, that's fair. I'm fine adding that text then.

README.md
```diff
@@ -40,3 +40,21 @@ Contributions are very welcome, as are feature requests and suggestions. Please
 [codecov-url]: https://codecov.io/gh/JuliaData/CSV.jl
 
 [issues-url]: https://github.com/JuliaData/CSV.jl/issues
+
+## Alternatives
+
+There are several other packages for reading CSV files in Julia, which may suit your needs better:
+
+* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/).
```
nalimilan, line 48: DelimitedFiles is going to be removed from the stdlib soon, so probably better anticipate the wording and say "used to be in the stdlib": https://github.com/JuliaLang/julia/pull/45540.

mcabbott: Maybe this should wait until it can mention a version number? Can't be before 1.9, and since 1.8 is not yet released.

README.md
davidanthoff: I think I wouldn't mention the IterableTables.jl story, because one should be able to materialize CSVFiles into any Tables.jl sink as well. The main difference here is really just that CSVFiles.jl uses a different parsing code than CSV.jl.

davidanthoff:
> I'd prefer we remove the reference to CSVFiles.jl/TextParse.jl; they haven't been updated or had any real work done for a long time, and they don't provide any kind of functionality not supported in CSV.jl.

I'm fine either way, but it is probably worth pointing out that this set of packages works well and for some applications not having updates frequently is a huge plus. We are using these packages in my research group extensively, because they work just fine, and not having to deal with updates is a plus for many of our use-cases. And neither package is deprecated.

README.md
quinnj: I think it could be worth mentioned the FileIO.jl integration, since I think some people prefer the `load`/`save` api.

mcabbott: Perhaps:
```suggestion
* [CSVFiles.jl](https://github.com/queryverse/CSVFiles.jl) and uses the [FileIO.jl](https://github.com/JuliaIO/FileIO.jl) API,
  but otherwise has similar goals. Like CSV.jl, it works with [Tables.jl](https://github.com/JuliaData/Tables.jl objects.
```

README.md
```diff
@@ -40,3 +40,21 @@ Contributions are very welcome, as are feature requests and suggestions. Please
...
+  This returns a `Matrix` rather than a [Tables.jl](https://github.com/JuliaData/Tables.jl)-style container, thus works best for files of homogenous element type. 
+  On large files, CSV.jl will be much faster.
+
+* [CSVFiles.jl](https://github.com/queryverse/CSVFiles.jl) uses the [FileIO.jl](https://github.com/JuliaIO/FileIO.jl) API 
+  into any [IterableTables.jl](https://github.com/queryverse/IterableTables.jl) sink.
+  The package uses [TextParse.jl](https://github.com/queryverse/TextParse.jl) for parsing.
+
+* [DLMReader.jl](https://github.com/sl-solution/DLMReader.jl) also aims to be fast for large files,
+  closely associated with [InMemoryDatasets.jl](https://github.com/sl-solution/InMemoryDatasets.jl).
```
mcabbott, line 57: > DLMReader.jl is a great reference because it supports some more exotic parsing configurations if you need to get really custom.

Should this link say something like "exotic custom parsing"?

README.md
mcabbott:
```suggestion
* The standard library contains [DelimitedFiles.jl](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/), at least until Julia 1.8.
```

quinnj (approved):

PR #1018, opened 2022-08-30
Add SnoopPrecompile

jakobnissen:
On my computer, this reduces the time to first task by about 15-25%, depending
on the task at hand.

@quinnj I don't know the codebase, and so I'm not confident the workload I have in src/workload.jl is representative.
Using `@snoopi_deep` from SnoopCompile on `CSV.read(path, DataFrame)` reveals quite a lot of inference is still left, so it is likely possible that a workload file that exercises more functionality will lead to even better precompilation.


jakobnissen:
Though see also #975 - if there is excessive type instability in this package, precompilation may not be a good idea.

quinnj (approved):
Yeah, I think this is good. As mentioned, I think we have some additional cleanup to really get the most value out of this, but I think we merge this now and we can address other issues later.

PR #1023, opened 2022-09-09
Check for user-given type when encountering extra columns during parsing

### LINKED ISSUE #1021: Bug with setting column type
Reproducer:
```
julia> str = """
       1 2 3
       1 2
       1 2 3 4
       1
       1 2 3 4 5
       """
"1 2 3\n1 2\n1 2 3 4\n1\n1 2 3 4 5\n"

julia> CSV.read(IOBuffer(str), DataFrame, delim=" ", header=false, types=String, silencewarnings=true)
5×5 DataFrame
 Row │ Column1  Column2  Column3  Column4  Column5
     │ String   String?  String?  Int64?   Int64?
─────┼─────────────────────────────────────────────
   1 │ 1        2        3        missing  missing
   2 │ 1        2        missing  missing  missing
   3 │ 1        2        3              4  missing
   4 │ 1        missing  missing  missing  missing
   5 │ 1        2        3              4        5
```
As you can see columns 4 and 5 get an incorrect column type.

nickrobinson251:
- close #1021 
- Before this PR, when we encounter a new/extra column we were always relying on type detection -- by setting this extra column's type to `NeedsTypeDetection` -- or, if streaming, setting it to a string type. This meant we were ignoring the case where a user had actually provided the type for this "extra" column, e.g. by setting `types=T`, to request all columns to have eltype `T`.
- This PR largely factors out the Column-type setting logic used when initalising columns in `Context` to be available to be reused by `File` when a new/extra column is encountered during parsing.

### DIFF
--- a/src/context.jl
+++ b/src/context.jl
@@ -93,6 +93,55 @@ end
 
 @noinline nonconcretetypes(types) = throw(ArgumentError("Non-concrete types passed in `types` keyword argument, please provide concrete types for columns: $types"))
 
+function initialize_columns(ncols::Int, names, types, stringtype, streaming::Bool, validate::Bool, options::Parsers.Options)
+    customtypes = Tuple{}
+    if types isa AbstractVector
+        length(types) == ncols || throw(ArgumentError("provided `types::AbstractVector` keyword argument doesn't match detected # of columns: `$(length(types)) != $ncols`"))
+        columns = Vector{Column}(undef, ncols)
+        for i = 1:ncols
+            col = Column(types[i], options)
+            columns[i] = col
+            if nonstandardtype(col.type) !== Union{}
+                customtypes = tupcat(customtypes, nonstandardtype(col.type))
+            end
+        end
+    elseif types isa AbstractDict
+        T = streaming ? Union{stringtype, Missing} : NeedsTypeDetection
+        columns = Vector{Column}(undef, ncols)
+        for i = 1:ncols
+            S = getordefault(types, names[i], i, T)
+            col = Column(S, options)
+            columns[i] = col
+            if nonstandardtype(col.type) !== Union{}
+                customtypes = tupcat(customtypes, nonstandardtype(col.type))
+            end
+        end
+        validate && checkinvalidcolumns(types, "types", ncols, names)
+    elseif types isa Function
+        defaultT = streaming ? Union{stringtype, Missing} : NeedsTypeDetection
+        columns = Vector{Column}(undef, ncols)
+        for i = 1:ncols
+            T = something(types(i, names[i]), defaultT)
+            col = Column(T, options)
+            columns[i] = col
+            if nonstandardtype(col.type) !== Union{}
+                customtypes = tupcat(customtypes, nonstandardtype(col.type))
+            end
+        end
+    else
+        T = types === nothing ? (streaming ? Union{stringtype, Missing} : NeedsTypeDetection) : types
+        if nonstandardtype(T) !== Union{}
+            customtypes = tupcat(customtypes, nonstandardtype(T))
+        end
+        columns = Vector{Column}(undef, ncols)
+        for i = 1:ncols
+            col = Column(T, options)
+            columns[i] = col
+        end
+    end
+    return columns, customtypes
+end
+
 struct Context
     transpose::Bool
     name::String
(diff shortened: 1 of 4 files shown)


src/context.jl
```diff
@@ -79,19 +79,79 @@ function checkvaliddelim(delim)
...
+
+function initialize_column(i, types::AbstractDict, names, stringtype, streaming::Bool, options)
+    defaultT = streaming ? Union{stringtype, Missing} : NeedsTypeDetection
+    T = getordefault(types, names[i], i, defaultT)
+    col = Column(T, options)
+    return col
+end
+
+function initialize_column(i, types::Function, names, stringtype, streaming::Bool, options)
+    defaultT = streaming ? Union{stringtype, Missing} : NeedsTypeDetection
```
nickrobinson251, line 141: this logic is repeated in a few of these methods, so i considered pulling it out but that actually leads to it being repeated a few places further apart in the code which seemed less nice than it being contained only in the methods of this one function

src/context.jl
```diff
@@ -119,6 +179,11 @@ struct Context
     debug::Bool
     tempfile::Union{String, Nothing}
     streaming::Bool
+    types::Union{Nothing, Type, AbstractVector, AbstractDict, Function}
```
nickrobinson251, line 182: the least nice thing about this PR, in my view, is that (i) it requires yet-another field in `Context` and (ii) one with this long-winded type `Union{Nothing, Type, AbstractVector, AbstractDict, Function}`

On (i), i couldn't think of another way to pass this info through to the parsing context... it's kinda what `Context` is for.

On (ii), i considered whether when constructing the `Context` we could construct a new `types` with a fixed type e.g. `types::Vector`, but i think that actually becomes more tricky, since we always want the same flexibility with `types` whether we're using it _before_ parsing (like in Context) or _during_ parsing (like when we hit an extra column), so i decided against this.

Thoughts welcome!

src/file.jl
```diff
@@ -684,17 +684,18 @@ Base.@propagate_inbounds function parserow(startpos, row, numwarnings, ctx::Cont
                 # extra columns on this row, let's widen
                 ctx.silencewarnings || toomanycolumns(ncols, rowoffset + row)
                 j = i + 1
-                T = ctx.streaming ? Union{ctx.stringtype, Missing} : NeedsTypeDetection
                 while pos <= len && !Parsers.newline(code)
-                    col = Column(T, ctx.options)
+                    col = initialize_column(j, ctx)
                     col.anymissing = ctx.streaming || rowoffset == 0 && row > 1 # assume all previous rows were missing
                     col.pool = ctx.pool
+                    # TODO: Do we need to check `nonstandardtype(T)` and potentially create a `Context` an updated `customtypes`?
```
nickrobinson251, line 691: thoughts, @quinnj ?

src/context.jl
quinnj: This is probably fine. Still reviewing the rest of the PR

src/file.jl
quinnj: Oof, this is rough. This is exposing a big design issue w/ how we're currently doing parsing. And it's not obvious how to fix. Unfortunately just creating a new context isn't quite enough because we're cheating a little and passing `ctx.customtypes` to `parsechunk!` so that it gets specialized. So we would need to recall `parsechunk` _and_ pass the new customtypes, and my brain can't quite wrap around how this would affect the multithreaded context we're in.

BUT, I think this fix is good for now, but we maybe just need to make a note that non-standard types aren't supported on widened columns.

quinnj (approved):

quinnj:
Yeah, this looks good overall; but I think as we do the larger refactor with @Drvi's work, we should think more broadly about how to address the non-standard type problem and if we can come up with something clever there.

src/file.jl
nickrobinson251: ah, yeah this is complex -- i agree it's not obvious how to support it and it does seem _very_ niche.

I've added an `@error` log to inform us if we are ever in this scenario and tell a user parsing will probably fail
https://github.com/JuliaData/CSV.jl/pull/1023/commits/e00fe5e9fcd26c7076280ad7b69e2d519040012f

quinnj:
Looks like there's a real test failure happening during column widening.

nickrobinson251:
Glad i waited for Ci! Issues should be fixed now.

PR #1046, opened 2022-10-25
Clear thread state to ensure threads local state don't keep references

quinnj:
We're surely running into https://github.com/JuliaLang/julia/issues/40626.

This is my attempt to clear thread's local storage after multithreaded parsing.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -620,8 +620,16 @@ end
 # as suggested in the above issue, spawned tasks may
 # end up getting stuck in thread local storage
 # running clear_thread_states clears out any thread local storage tasks
+struct _Returns{V} <: Function
+    value::V
+    _Returns{V}(value) where {V} = new{V}(value)
+    _Returns(value) = new{_stable_typeof(value)}(value)
+end
+
+(obj::_Returns)(@nospecialize(args...); @nospecialize(kw...)) = obj.value
+
 function clear_thread_states()
     Threads.@threads :static for _ in 1:Threads.nthreads()
-        Timer(Returns(nothing), 0; interval = 1)
+        Timer(_Returns(nothing), 0; interval = 1)
     end
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -622,8 +622,6 @@ end
 # running clear_thread_states clears out any thread local storage tasks
 struct _Returns{V} <: Function
     value::V
-    _Returns{V}(value) where {V} = new{V}(value)
-    _Returns(value) = new{_stable_typeof(value)}(value)
 end
 
 (obj::_Returns)(@nospecialize(args...); @nospecialize(kw...)) = obj.value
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -614,4 +614,20 @@ macro refargs(ex)
...
     end
-end
\ No newline at end of file
+end
+
+# https://github.com/JuliaLang/julia/issues/40626
+# as suggested in the above issue, spawned tasks may
+# end up getting stuck in thread local storage
+# running clear_thread_states clears out any thread local storage tasks
+struct _Returns{V} <: Function
```
bkamins, line 623: why do you define `_Returns` instead of using `Base.Returns`?

quinnj: I ran into compat issues with 1.6; maybe I just needed to do `Base.Returns` instead and it was exported starting in 1.7?

src/utils.jl
```diff
@@ -614,4 +614,20 @@ macro refargs(ex)
...
+    value::V
+end
+
+(obj::_Returns)(@nospecialize(args...); @nospecialize(kw...)) = obj.value
+
+function clear_thread_states()
```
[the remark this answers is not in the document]
quinnj: It's detailed in the https://github.com/JuliaLang/julia/issues/40626 issue. Each thread basically has a thread local storage where they will store the most recent Task that it ran, which includes a reference to the _result_ of the task that ran. So to ensure each thread's local state is cleared, we run a dummy no-op task on each thread (via `:static`) to ensure no references remain to items that should be candidates for GC.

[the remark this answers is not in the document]
bkamins: Thank you!

src/utils.jl
bkamins: Ah yes, docs states:

> Returns requires at least Julia 1.7.

iamed2:
I believe this broke multithreaded parsing within `asyncmap` on Julia 1.6 (might affect other versions). 

You can replicate by doing any `asyncmap` that calls a `CSV.read` that uses the multithreaded parsing, e.g.:

```julia
asyncmap(x -> CSV.read("hw_25000.csv", DataFrame; ntasks=3), 1:3; ntasks=3)
```

quinnj:
@iamed2, would you mind trying out [this PR](https://github.com/JuliaData/CSV.jl/pull/1050); hopefully that is a little more conservative in clearing thread states.

iamed2:
Looks like that fixed it! Thanks!

Btw I just pulled the CSV file from https://people.sc.fsu.edu/~jburkardt/data/csv/csv.html; my example above should be fully replicable. I originally demonstrated the issue using a bare `julia:1.6` docker image with just CSV+DataFrames installed. 

Might be worth adding a test. Although I don't think it's likely this exact problem with reoccur, I don't see any tests that involve `@async`, `asyncmap`, or nested tasks, and that's both a very common use case for us across several private packages.

quinnj:
Ah that makes sense. Thanks for confirming. Yeah, I'll look into adding a test.

Issue #1057, opened 2022-12-04
Design of clear_task_threads


bkamins:
In https://github.com/JuliaData/CSV.jl/blob/35e840254ff50de2211a5cbbc02924f842b241bb/src/utils.jl#L629 we have:
```
function clear_thread_states()
    # only clear thread states in the workflow where a user is running
    # CSV.File from the REPL at the top-level; if we're already spawned
    # in a task, we don't want to mess up thread task states
    if current_task() == Base.roottask
        Threads.@threads :static for _ in 1:Threads.nthreads()
            Timer(_Returns(nothing), 0; interval = 1)
        end
    end
end
```

@quinnj I do not understand two things:
* why this gets executed if we are in root task only? The same problem can be if we are not in root task;
* I do not understand why we use `Timer` here? I understand why it is proposed in https://github.com/JuliaLang/julia/issues/40626 (as it would be a solution in general for Base Julia) but why call this function every second forever from CSV.jl? I would think it is enough to call it once.

quinnj:
Ok, I've dug back into this and here's my understanding:
  * There are 2 ways that references to things are kept around unexpectedly:
    * By being returned from a `Threads.@spawn` call; even if the object is not referenced in user code, the created `Task` will keep a reference to it as what was returned from running the task
    * Any inputs that get captured in the closure created to run a `Threads.@spawn` call; again, the `Task` keeps the reference to these, even if they aren't further referenced in user code
  * Because of the 2 ways to keep references above, once a Task has finished running, each physical thread keeps a reference to its "current task", which includes its most recently run Task if another has not yet been scheduled; by keeping a reference to a `Task`, it keeps references to the closure inputs and returned results as mentioned above

One easy mitigation, if possible, is to ensure that `Threads.@spawn` calls return `nothing`. That avoids at least one case of a reference to something being kept around unexpectedly.

The closure inputs reference, however, is unavoidable as far as I understand. This leads us to consider alternative solutions, like https://github.com/JuliaLang/julia/pull/47405, which basically tries to treat Tasks that have finished as weak refs, or the `clear_thread_states` as we've attempted in CSV.jl.

The original definition in CSV.jl was:

```julia
function clear_thread_states()
    Threads.@threads :static for _ in 1:Threads.nthreads()
            Timer(_Returns(nothing), 0; interval = 1)
        end
end
```

[...]

PR #1058, opened 2022-12-07
Use WeakRefs for spawned tasks to avoid holding unexpected references

quinnj:
Most complete explanation is [here](https://github.com/JuliaLang/julia/issues/40626#issuecomment-1340376861).

Also discussed [here](https://github.com/JuliaData/CSV.jl/issues/1057).

This PR proposes an alternative solution to `clear_thread_states` where that approach can be problematic (interferring with global thread state, not working as expected in nested spawned tasks, etc.). The previous definition also started unending `Timer` tasks that could build up over time.

The approach in this PR is to wrap spawned task closure arguments in `WeakRef` to allow them to be GCed as expected once the tasks are finished.


src/utils.jl
```diff
@@ -616,23 +616,45 @@ macro refargs(ex)
...
+
+A macro that wraps a `Threads.@spawn` block with `WeakRef` calls for all `args...` arguments, allowing
+them to be garbage collected once the `Task` has finished running. Must be used within a `@syncpreserve`
+block to ensure input arguments are not garbage collected for the lifetime of the `@sync` block (since
+otherwise they would be allowed to be garbage collected before the `Task` finishes running).
+Note that to use `args...` inside your `Threads.@spawn` block, you must use the same argument name,
+but with an `_` prefix; this avoids the original input being capture by `Task` closure. So usage is like:
+
+```julia
+@weakrefspawn ctx foo begin
```
bkamins, line 635: maybe add the `@sync` part around it to give a full example of intended use?

src/utils.jl
```diff
@@ -616,23 +616,45 @@ macro refargs(ex)
...
+unwrap(val) = val
+unwrap(val::WeakRef) = val.value
+
+"""
+    @weakrefspawn args... begin
+
+    end
+
+A macro that wraps a `Threads.@spawn` block with `WeakRef` calls for all `args...` arguments, allowing
+them to be garbage collected once the `Task` has finished running. Must be used within a `@syncpreserve`
```
bkamins, line 628: Thank you for working on this!

Above `@sync` not `@syncpreserve` is used. Which is correct?

Also are we sure that the "preserve" part touches only the selected variables (this is relevant, as inside the task GC might have to be run if the taks itself allocates a lot).

quinnj:
@bkamins, I've updated this PR to use the `@wkspawn` macro now provided in the WorkerUtilities.jl package: https://github.com/JuliaServices/WorkerUtilities.jl/pull/10. I moved it there because I have a few other multithreaded/concurrency-related utilities there and I want to use this in the Arrow.jl package as well. 

As noted in that PR, the `@wkspawn` macro is a drop-in replacement for `Threads.@spawn`, but will pass mutable closure arguments wrapped in `WeakRef` to the spawn block expression.

PR #1060, opened 2022-12-13
Also precompile CSV.File(::String)

timholy:
A tiny enhancement that will be more noticeable with an ~upcoming~ [Julia PR](https://github.com/JuliaLang/julia/pull/47889).


quinnj (approved):
Thanks @timholy!

Issue #1061, opened 2022-12-14
Invalidations when loading DataFrames


timholy:
Some non-inferrable code in CSV gets invalidated when you load DataFrames. This is a sister issue to https://github.com/JuliaData/DataFrames.jl/issues/3248, and you can read that issue for a more complete discussion of potential ways to address the issue. In CSV's case, the only invalidations you need to care about come from DataStructures; similar to the suggestion in https://github.com/JuliaData/DataFrames.jl/issues/3248, the simplest approach might be to make DataStructures an obligate part of your package. But fixing the inference failure(s) (I haven't tried) would avoid the invalidation without taking on DataStructures as a dependency.

```julia
julia> using SnoopCompileCore

julia> invs = @snoopr begin
           using DataFrames
           using CSV
       end;

julia> tinf = @snoopi_deep CSV.File(joinpath(pkgdir(CSV), "test", "testfiles", "precompile.csv"));

julia> using SnoopCompile

julia> tinf
InferenceTimingNode: 7.241082/7.613602 on Core.Compiler.Timings.ROOT() with 15 direct children

julia> trees = invalidation_trees(invs);

julia> staletrees = precompile_blockers(trees, tinf)
1-element Vector{SnoopCompile.StaleTree}:
 inserting iterate(v::Union{Base.KeySet{<:Any, <:DataStructures.SwissDict}, Base.ValueIterator{<:DataStructures.SwissDict}}, state) @ DataStructures ~/.julia/packages/DataStructures/59MD0/src/swiss_dict.jl:646 invalidated:
[...]

PR #1069, opened 2023-01-15
`typemap`: switch to IdDict

timholy:
When using types as keys, IdDict is essentially always preferred:
- it uses `@nospecialize` on all operations, reducing latency
- it uses pointer comparisons to test for matches, which is much faster than subtyping (which is what `isequal` dispatches to).

The second commit adds a lot of noise in the form of whitespace fixes.
Feel free to drop this commit if you prefer (and easiest to review by
suppressing whitespace changes).

This fixes some invalidations (see #1061)


quinnj (approved):
Thanks! Sounds like an overall win, and even for the API, it makes sense to use.

PR #1073, opened 2023-02-12
Some fixes for multithreaded Context parsing

### LINKED ISSUE #1010: Threaded parsing type mismatch depending on `ntasks` value
## Problem

We've ran across a very odd issue where depending on the value of `ntasks` set the type of parsed is different.

```julia
CSV.File("foobar.csv"; ntasks=60, debug=true)
...
types after parsing: Type[Float64], pool = (0.2, 500)
```

```julia
CSV.File("foobar.csv"; ntasks=120, debug=true)
types after parsing: Type[String31], pool = (0.2, 500)
```

File to replicate the issue, [foobar.csv](https://github.com/JuliaData/CSV.jl/files/9003358/foobar.csv)

### LINKED ISSUE #1047: Integer column randomly typed as string when threading enabled
The following code:
```
using CSV
using DataFrames
using Random

NCOLS = 30
NROWS = 150
Random.seed!(1)
fname = tempname()
f = open(fname, "w")
write(f, join("col".*string.(1:NCOLS), ","))
write(f, "\r\n")
for i in 0:NROWS
	write(f, join(string.(rand(Int16, NCOLS)), ","))
	write(f, "\r\n")
end
close(f)
df_by_threads = CSV.read(fname, DataFrame)
df_single_threaded = CSV.read(fname, DataFrame; ntasks=1)
print(eltype.(eachcol(df_by_threads)) == eltype.(eachcol(df_single_threaded)))
```
will print `false` , or at least it does on my Windows box, with 8 threads, running CSV v0.10.7.

If I reduce NCOLS or NROWS it will be true. If I choose a different random seed it may become true.

If one inspects the columns of `df_by_threads`, at least one column will be of type `String7`, but which column may vary with repeated execution and sometimes there are two such columns, even though the data written to the file is fixed.

Liozou:
I had a CSV file with only `Float64` whose columns were sometimes parsed as `String`, sometimes as `Float64` when using multiple threads. Whether it happened and which columns were affected changed from one execution to the other, which made the issue difficult to track.

This PR fixes this behaviour and a couple of issues encountered in the same area of the code. Most notably:
- this fixes the race condition underlying the not-reproductible issue: this came from `findchunkrowstart`, where thread `i` reads `ranges[i+1]` and writes to `ranges[i]`.
- if multithreaded parsing fails, the column types are now properly reinitialized to allow the fallback single-threaded parsing to re-infer them correctly (I followed the suggestion from #1010).
- a trailing newline at the end of the file could be mistaken for a row with 0 columns, causing multi-threaded parsing to fail.

I also made multi-threading parsing failure a loud `@error` to prevent performance pitfalls to go unnoticed (last commit). This can be easily reverted if you think it can lead to too many spurious logs.

Fix #1010, fix #1047 and added tests (which now include one occurrence of the new `@error`).


Liozou:
Bump. This fixes a real bug and a race condition.
If the changes are too complex to review, would you like me to split this PR in smaller separate ones? I tried to do that with the different commits already, but I understand if it's still difficult to navigate.

src/detection.jl
```diff
@@ -336,24 +336,23 @@ ColumnProperties(T) = ColumnProperties(T, 0x00)
     end
 end
 
+function findnextnewline(pos, stop, buf, opts)
+    while pos < stop
```
quinnj, line 340: `stop` here seems to be the last byte of the current chunk; why is it `<` here instead of `<=`? I understand the change below from `len = ranges[i + 1]` to `len = ranges[i + 1] - 1`, but it seems like we still want to check each byte in the chunk?

quinnj: (for example, on line 372 we're doing `while pos <= len`...)

quinnj (commented):
This looks pretty good IMO; I left one question though. I'm also not sure why CI is not running at all here? We definitely want to make sure we have a good test run before merging.

src/detection.jl
Liozou: Thanks for the review!

This function `findnextnewline` only returns the position of the first encountered newline and defaults by returning `stop`. So, whether there is no newline, or whether the newline happens on the last byte, it will return `stop` either way and there is no side-effect to the function, so we might as well skip checking that last byte.
The default return value comes from the assumption that each chunk should end with a newline. Line 450 this assumption is verified by the preprocessing of ranges from lines 466-480, while on the call line 470 `stop` is simply the end of the buffer.

quinnj (approved):

quinnj:
Thanks @Liozou!

Liozou:
You're welcome, thanks for the review and merge! I believe this should also fix https://github.com/JuliaData/CSV.jl/issues/1089, I checked locally that the bug occurs when using e.g. `ntasks=8` but not after this PR.

PR #1078, opened 2023-04-20
Example updated to show conversion to a Matrix

a2ray:


quinnj (approved):
Thanks for the contribution!

Issue #1080, opened 2023-04-20
getproperty on File makes internal use of dot notation problematic


albheim:
From a [discourse thread](https://discourse.julialang.org/t/csv-jl-fails-when-reading-multiple-inputs-at-once-if-variable-name-name-occurs).

Since getproperty on `File` is defined to first look up if any field is in the internal lookup table
https://github.com/JuliaData/CSV.jl/blob/cfb4ffb5d9847df4f8d5efa5129b27026a83a4a8/src/file.jl#L142-L145
we get a problem when using getproperty (i.e. dot syntax) to access the fields of the struct, as can be seen here
https://github.com/JuliaData/CSV.jl/blob/cfb4ffb5d9847df4f8d5efa5129b27026a83a4a8/src/file.jl#L940

where if any of the column names from from the data read into `f` (which is type `File`) shares name with one of the fields that are accessed there, it will likely error or maybe do incorrect things at least.

In the linked thread this happened when one column had the name `name`, which then errored.

albheim:
Solved by #1081

PR #1081, opened 2023-04-20
Fix getproperty calls on fields for File

albheim:
Solves #1080 and adds test for the original instance of the problem.

Would it be worth having a check on all field names? Feels like it is very easy to accidentally use the dot syntax internally when accessing the fields, and only when someone use data where the columns share the names something would be incorrect. Maybe more than only this constructor should be tested with columns using the field names of `File`?


quinnj (approved):
Thanks for the fix!

Issue #1083, opened 2023-04-29
Too many missing warnings


adienes:
```
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 113. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 114. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 115. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 116. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 117. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 118. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 119. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 120. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 121. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
┌ Warning: thread = 1 warning: only found 9 / 11 columns around data row: 122. Filling remaining columns with `missing`
└ @ CSV ~/.julia/packages/CSV/b8ebJ/src/file.jl:576
```

one warning at most is sufficient. definitely not one per row. honestly I would expect this as default behavior so I'd be ok if it didn't warn at all

jariji:
I would rather get an error by default.

quinnj:
[...]

PR #1084, opened 2023-05-02
fewer default_max_warnings

adienes:
addresses https://github.com/JuliaData/CSV.jl/issues/1083


quinnj:
You can control this via the `maxwarnings` keyword arg, so I don't think this change is as necessary.

PR #1093, opened 2023-06-02
Support groupmark

### LINKED ISSUE #626: Ability to specify a thousands separator character
This feature was listed in https://github.com/JuliaData/CSV.jl/issues/3 , but not implemented.  

https://github.com/JuliaData/CSV.jl/issues/207 has a complicated workaround (but keyword transforms does not exists anymore?), and says "there isn't a "thousands" separator option, but we could potentially add on".

Practical example, data is two first rows from table 2 on page 10 in https://www.istat.it/it/files//2020/05/Rapporto_Istat_ISS.pdf
```
a=CSV.File(IOBuffer("""
       Alessandria 95,7 98,2 -12,8 91,0 1.199 693 222 18,5
       Ancona 76,6 84,3 -10,7 49,4 704 528 86 12,2
       """), delim=" ", header = false, decimal = ',')
```
Row 1, column 6 would ideally be Int 1199.   Here comma is a decimal separator and period is a thousands separator.

LilithHafner:
Closes #626. These features don't implement themselves, even when seven separate people comment asking for them.

See also https://github.com/JuliaData/Parsers.jl/issues/168. This PR fixes none of that.

This is my first time working with this codebase and its style and I may have missed something. I'd appreciate a careful review.


quinnj (approved):
This looks great @LilithHafner! Can I request one more change? We have the reading.md file where we try to spell out in detail all the various keyword args that are supported. Could you add an entry there for `groupmark` and if you're feeling ambitious, link to an example in the example.md file?

LilithHafner:
That's the kind of request that leads to well-documented packages. Thank you.

quinnj (approved):

LilithHafner:
That is unfortunate—I imagine the best fix is the stop mixing up those arguments rather than restricting compat. Unfortunately, I do not have the bandwidth to debug that rn. Perhaps @quinnj is willing to look into it?

PR #1099, opened 2023-06-17
Selectively reduce multithreaded parsing @error

Liozou:
Reduce the `@error` introduced in https://github.com/JuliaData/CSV.jl/pull/1073 to a `@warning` whenever previous warnings may explain why multithreaded parsing failed (typically, unequal number of columns per row). In those cases, it will print:
```julia
┌ Warning: Multithreaded parsing failed and fell back to single-threaded parsing, check previous warnings for possible reasons.
```

If there is no possible explanation, this still shows a red `@error` saying:
```julia
┌ Error: Multithreaded parsing failed and fell back to single-threaded parsing. This can happen if the input contains multi-line fields; otherwise, please report this issue.
```

In both cases, a debug statement is printed as soon as multithread parsing failed (if the Context `debug` kwarg is set), whereas the `@warning` or `@error` only appears at the end of the fallback single-threaded parsing.

I think this compromise allows us to still get error reports like https://github.com/JuliaData/CSV.jl/issues/1095, which are crucial to detect and allow use to fix underlying issues where multithreaded parsing should not fail (like https://github.com/JuliaData/CSV.jl/pull/1098 in this instance), while avoiding to stress the user with an `@error` when there is nothing to report.

Close https://github.com/JuliaData/CSV.jl/issues/1095.


quinnj (approved):
Thanks!

PR #1106, opened 2023-07-26
Fix the problem in Example page since v0.10

hungpham3112:
Not check yet, I'm not familiar with `Documenter.jl` and I can't run make.jl in my local machine so need to check before merge. Thanks.

v0.10.11:
image


quinnj (approved):
Thanks @hungpham3112!

hungpham3112:
Hi, I checked the document but still see the error. The document haven't update yet?

image

giordano:
@hungpham3112 there haven't been stable releases of CSV.jl after your PR, yet, but the fix is visible in the dev docs: https://csv.juliadata.org/dev/examples.html

quinnj:
I just kicked off a new patch release

PR #1121, opened 2024-02-06
Support arbitrary `DenseVector{UInt8}` as valid sources.

oscardssmith:
The specific motivation of this is so that in Julia 1.11, `Memory` is a valid source, but it does seem reasonable in general.


vtjnash:
The problem in that list is the existing type IO, not the added type: this claims to support all IO, but then has some special cases to handle GenericIOBuffer which specifically are incomplete (as it accesses `.data` directly rather than going through `take!` in the `else` block)

PR #1123, opened 2024-02-15
Update keyworddocs.jl for limit to remove use of deprecated "threaded"

rdavis120:
The ```threaded = false``` arg is deprecated and should be removed from the documentation of the limit arg.


quinnj (approved):
Thanks!

PR #1127, opened 2024-02-28
Update ci.yml: Add mac aarch64 CI, codecov v4

ViralBShah:
Add Mac aarch64 CI
Add codecov v4
Add a 20 min timeout
Run stable x86, LTS x64, and nightly x64 only on linux


ViralBShah:
We should merge this - but there seems to be an issue on precompilation with julia nightly.

quinnj (approved):
Thanks @ViralBShah!

PR #1133, opened 2024-04-09
Fix breakage caused by JuliaLang/julia/pull/53896

### LINKED ISSUE #1132: CSV is broken in nightly julia
https://github.com/JuliaLang/julia/pull/53896 removed `Base.wrap` and with that, the fix to #1124 . at least i´m now familiar on how it fails, fix incoming

longemen3000:
https://github.com/JuliaLang/julia/pull/53896 removed `Base.wrap`, and i did use that to fix CSV in the last patch. This now uses `view(x::Memory,len)` instead.

fixes #1132

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -245,10 +245,11 @@ end
 # one-liner suggested from ScottPJones
 consumeBOM(buf, pos) = (length(buf) >= 3 && buf[pos] == 0xef && buf[pos + 1] == 0xbb && buf[pos + 2] == 0xbf) ? pos + 3 : pos
 
-if isdefined(Base,:wrap)
-    __wrap(x,pos) = Base.wrap(Array,x,pos)
+if isdefined(Base, :Memory)
+    __wrap(x) = view(x,Base.OneTo(length(x)))
+    __wrap(x::Array) = x
 else
-    __wrap(x,pos) = x
+    __wrap(x) = x
 end
 
 # whatever input is given, turn it into an AbstractVector{UInt8} we can parse with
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -245,10 +245,11 @@ end
 # one-liner suggested from ScottPJones
 consumeBOM(buf, pos) = (length(buf) >= 3 && buf[pos] == 0xef && buf[pos + 1] == 0xbb && buf[pos + 2] == 0xbf) ? pos + 3 : pos
 
-if isdefined(Base,:wrap)
-    __wrap(x,pos) = Base.wrap(Array,x,pos)
+if isdefined(Base, :Memory)
```
oscardssmith, line 248: can't you get rid of this entire function and just make it use `view`

longemen3000: i want `view(x::Array)` return `x`, and `view(x::Memory)` does not work

longemen3000: or do you mean in the function `getbytebuffer` itself?

oscardssmith: yeah, this function only has one use and now that view isn't a new name from base you can just use it without the check.

Keno:
I'd like to go ahead and merge this to fix the regression on master. If there's additional simplification to be had, let's do a follow up PR.

src/utils.jl
topolarity, line 248: Julia 1.11 also has Base.Memory (but not the recent `wrap` rename) - Will this PR be backwards compatible?

oscardssmith: the `wrap`->`view` will be back-ported to 1.11

Keno: It seems rude to break people on the current RC. Can we put in appropriate checks?

longemen3000: I can add an extra if for the case that `Memory` and `wrap` is defined

Keno (approved):

ViralBShah:
I'm merging to fix the breakage - and perhaps further improvements can follow.

PR #1158, opened 2025-03-18
Update examples.md to use ZipArchives

nhz2:
I don't think ZipFile.jl is being maintained anymore, so I am updating the example to use ZipArchives.jl

This should also fix @JakeZw 's issue in https://github.com/fhs/ZipFile.jl/issues/109

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -106,7 +106,7 @@ file = CSV.File(http_response)
 ## [Reading from a zip file](@id zip_example)
 
 ```julia
-using ZipFile, CSV, DataFrames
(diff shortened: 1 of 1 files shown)


quinnj (approved):
Thanks @nhz2!

JakeZw:
When I review the updated example, I fail to find where the file has been closed after the `ZipReader` command.  ZipArchive says it needs to be closed in https://docs.juliahub.com/General/ZipArchives/stable/autodocs/  under the 
ZipArchives.zip_openentry entry, the reading is not using the do block format.  

Am I missing something?

PR #1166, opened 2025-08-21
Improve zip file example

nhz2:
As noted by @JakeZw in https://github.com/JuliaData/CSV.jl/pull/1158#issuecomment-3189621162

The zip file example was not properly closing things.
And as noted in https://github.com/JuliaIO/ZipArchives.jl/issues/90, using mmap might not always be ideal.

This PR changes the example to avoid using streaming or mmap and instead buffers everything in memory to keep things simple.

I can give a more complete example based on the code in https://github.com/JuliaIO/ZipArchives.jl/blob/v2.5.1/test/test_file-array.jl
But this can get more complicated.


quinnj:
[Reviewed by AI] Scoped review of the example at `8cf0b96`.

Both zip-writing variants round-trip correctly through the proposed `ZipReader(read(...))` / `zip_readentry` path with ZipArchives 2.5.1, CSV main `ff0c032`, and Julia 1.12.6. The shell-mode line was run as `run(...)` in the script.

Current main already closes the archive file through `open(mmap, "a.zip")` and uses `zip_readentry`. The remaining change is the choice to buffer the compressed archive with `read`. Keep the memory note: this example holds the archive bytes and the uncompressed entry bytes. Please port that small change onto current docs; this PR currently has 154 changed files and merge conflicts. I did not test the full historical PR checkout.

PR #1170, opened 2025-10-03
Fix compatibility with Julia 1.13+ memhash removal

vtjnash:
Remove hash method definition when Base.memhash is not available.
On Julia 1.13+, these AbstractString types will use the default
AbstractString hash implementation which is now efficient and
zero-copy based on codeunit/iterate.

For Julia <1.13, continue using the memhash-based implementation
for compatibility.

Related to JuliaLang/julia#59697

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -578,9 +578,11 @@ struct PointerString
     len::Int
 end
 
-function Base.hash(s::PointerString, h::UInt)
-    h += Base.memhash_seed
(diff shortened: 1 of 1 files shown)


vtjnash:
Rebased on forced pushed main

PR #1192, opened 2026-07-24
Keep source column when sources vector has one file

### LINKED ISSUE #1146: source column-name missing when the input vector contains only one csv-file
When `files` contains more than one csv-file name, the resulting table includes as expected and [documented](https://csv.juliadata.org/stable/reading.html#CSV.File) the chosen `source` column name:
```julia
tbl = CSV.File(files; source = :source)
@assert length(files) > 1
:source ∈ keys(first(tbl)) # true
```
But when the `Vector` `files` contains only one item, the resulting table does not.
```julia
tbl = CSV.File(files; source = :source)
@assert length(files) == 1
:source ∈ keys(first(tbl)) # false
```

JohnCobbler:
`File(sources::Vector; source=...)` short-circuits a one-element vector to
`File(sources[1]; kw...)`, forwarding only `kw...`. The `source` keyword is bound
to its own parameter, so it is not part of `kw` and gets dropped - a single-element
`sources` vector silently loses the requested source column, while the source-column
injection only runs in the multi-source path.

Skip the short-circuit when `source` is set, so a one-element vector falls through
to the general path that adds the column. The `source === nothing` fast path is
unchanged.

Added regression tests for a single-element `sources` vector: the `source` column
is present with `Symbol`/`String`/`Pair` names, and the no-`source` case still
matches the scalar `CSV.File(path)` result.

Fixes #1146

### DIFF
--- a/src/file.jl
+++ b/src/file.jl
@@ -912,7 +912,7 @@ function File(sources::Vector;
     if source isa Pair
         length(source.second) == length(sources) || throw(ArgumentError("source pair keyword argument list ($(length(source.second))) must match length of input vector ($(length(sources)))"))
     end
-    length(sources) == 1 && return File(sources[1]; kw...)
+    source === nothing && length(sources) == 1 && return File(sources[1]; kw...)
     all(x -> x isa ValidSources, sources) || throw(ArgumentError("all provided sources must be one of: `$ValidSources`"))
     kws = merge(values(kw), (ntasks=1,))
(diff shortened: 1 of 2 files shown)


quinnj:
Sorry for the slow review; these changes are superseded by the [1.0 release](https://github.com/JuliaData/CSV.jl/pull/1196). [issue resolution reviewed by AI; please respond with objections or additional feedback/concerns/issues]

Issue #1048, opened 2022-10-30
CSV.Chunks memory leak


bkamins:
Code used (on 4 threads):
```
julia> using CSV, DataFrames

julia> tables = CSV.Chunks("large.csv", ntasks=100);

julia> for t in tables
           df = DataFrame(t)
       end
```

The result is that huge amounts of RAM are allocated by Julia process and never reclaimed, even by `GC.gc()`.
Using `clear_thread_states()` from https://github.com/JuliaData/CSV.jl/pull/1046 does not resolve the issue.
The expectation is that I use `CSV.Chunks` to read very large files (larger than RAM), which is effectively made problematic with this problem.

quinnj:
I tried to reproduce this against both `0.10.16` and the new `1.0.0`, on a synthetic file. Short version: **there is no unbounded leak in either version**, the 0.10.x symptom is fully explained by four bounded effects, 1.0 fixes the part that mattered — but 1.0 introduces a *new* cost that still blocks the actual goal in the issue (files larger than RAM). Details below.

Setup: macOS arm64, 96 GiB RAM, Julia 1.12.6 and 1.8.5, `JULIA_NUM_THREADS=4`. Files: `big.csv` 2.72 GiB / 30M rows x 10 mixed columns, `half.csv` = its first half, `narrow.csv` 0.149 GiB / 40M rows x 2 small Int columns. The loop is exactly the one reported: `for t in CSV.Chunks(f; ntasks=100); df = DataFrame(t); end`.

### 1. Not a leak

Running the whole sweep 3x in one process, with a full `GC.gc(true)` between rounds (CSV 0.10.16, `ntasks=4`, `big.csv`):

```
round 1 done  rss=3.338 GiB  live=0.731 GiB  peak=6.062 GiB
round 2 done  rss=3.507 GiB  live=0.733 GiB  peak=6.232 GiB
round 3 done  rss=3.869 GiB  live=0.726 GiB  peak=6.593 GiB
```

`Base.gc_live_bytes()` is flat across rounds, so nothing accumulates. Julia 1.8.5 behaves the same as 1.12.6, so this is not a Julia-version story either.

### 2. What the 0.10.x symptom actually was

[...]

PR #1204, opened 2026-09-15
Stream CSV.Chunks over windows of the source

quinnj:
`CSV.Chunks` no longer builds a structural index of the whole source. One batch is one window of the source's data bytes: the window is indexed when its batch is produced and released with it. The index of a narrow numeric file is several times the size of the file (issue #1048), so a chunked read of a file larger than memory could not start in 1.0.0.

This replaces the earlier approach on this branch (a release/rebuild lifecycle on `ChunkIndex`), which indexed the source three times and threaded index-presence checks through every reader.

## Design

- A **window** is a byte range of the data region that starts and ends at a row boundary. It is the streaming unit and the batch. `chunkplan`/`index` take a `stop` byte and a `windowbytes` size; `BufferIndex.nextstart` says where the next window begins.
- A **`ChunkIndex`** stays the parallel work unit inside one window. It is built for the window, used, and dropped. Nothing releases, rebuilds, or mutates an index after it is built.
- `_prepare` no longer indexes the data range. Names come from a small header-prefix index; the data byte range comes from the existing index-free row walks (`_rawrowoffset`, `_footeroffset`). `File`, `lazy`, `Rows`, and `Tables.Scan` index the range once through `_prepareindexed` and run as before; their output is byte-identical.
- A batch is the existing eager driver `_parse` over the window's index with the settled schema as a requested-type plan, followed by the same finishing steps as `File`. Batches own their bytes.
- The schema pre-pass streams the same windows before the first batch, so every batch keeps the 1.0 guarantee: one column type, missingness, and settled text width. That pass also counts the batches, so `length` is still defined.

## Semantics

- `chunkbytes` is the target batch size in source bytes. It defaults to 64 MiB, or the whole source when it is smaller. One complete row can push a batch past it.
- `ntasks` bounds the parallel work inside a window. It is not a batch count and no longer affects the batch size.
- Batch boundaries differ from 1.0.0 for the same options. Rows, column types, values, and diagnostics do not. Problems carry file-global row numbers across batches; `on_error=:warn` warns once.
- Only the structural index is bounded. The source bytes stay in memory for later batches (memory-mapped for a large local file); a returned batch owns its own bytes.

## Validation

- Test suite: 331,543 passes at `-t8` on Julia 1.12.6 (`--check-bounds=yes`), 331,097 on Julia 1.10.11; `-t1` runs of `test/api.jl` and `test/scan.jl` pass with the same counts.
- `bench/fingerprint.jl`: every `File`/`lazy`/`Scan`/`Rows`/`write` fingerprint identical to `main`; only the 22 per-batch `chunks` lines differ (batch boundaries). Concatenated batch values, row counts, and schemas equal the `CSV.File` read for 22 shapes x 5 window sizes.
- Differential harness against `main` (`header` x `skipto` x `footerskip` x `limit` x `comment` x `chunkbytes` over 25 inputs with comment rows, empty rows, bare quotes in the head/data/footer, an unclosed quote at EOF, CRLF, BOM, aligned-space columns): 126,000 cases comparing names, eltypes, values, row counts, problem sets, and thrown errors for `File`, `Rows`, and `Chunks`. Every mismatch is on the two aligned-space inputs and is the documented fix below; on those inputs `File` and `Chunks` agree with each other at every batch size (84,000 comparisons, 0 disagreements).
- `bench/hygiene.jl` identical to `main`.
- Memory held by a `Chunks` object after construction: narrow 215 MiB file (20M x 2 Int) 327 MiB on `main`, 0 here; mixed 275 MiB file (3M x 10) 168 MiB on `main`, 0 here.
- Time, build + iterate, 8 threads, interleaved with `main`: narrow 0.39 s -> 0.33 s; mixed 0.23 s -> 0.26 s at the 64 MiB default (0.22 s at `chunkbytes=128 MiB`). The mixed-file gap is all in iteration: each window is indexed again when its batch is produced, so the source is indexed twice; on a narrow file not holding the whole index more than pays for that.

## One intentional fix

A quote that does not start its field makes a reader prepare the source again under the lenient quote rule. In 1.0.0 that retry kept the sniffed delimiter but dropped a sniffed `ignorerepeated`, so an aligned-space file with a bare quote gained an empty column (`[:a, :Column2, :b]`). `CSV.File` always hit the retry; a windowed `Chunks` with `limit` might not, so the schema would have depended on `chunkbytes`. The retry now freezes both, which makes the retry independent of where it is triggered. Documented in the new 1.1 release-notes section.

## Follow-ups (not in this PR)

- `CSV.Rows` still indexes the whole row window. A per-window `Rows` works with a sticky lenient retry, but it conflicts with the allocation contract pinned by #1203; separate change.
- The default window size is a knob: 128 MiB removes the mixed-file gap at the cost of a larger transient index.
- Preserved 1.0.0 behavior, not fixed: a listed multi-row header restores its cursor, so a row consumed as a header part can appear as data again when it sits at or after the last listed row.

Addresses the retained-index part of #1048.

Co-authored by Codex (three review rounds with independent reproductions).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


quinnj:
[Reviewed by AI]

Reviewed window boundaries, schema settling, row-limit/footer handling, and global diagnostic offsets at `8e7c8ac`. No blocking finding in this scope. The full Julia 1.12.6 suite passed 331,543 checks with bounds checking and four threads; Aqua passed 11 checks. An additional 6,000 deterministic File-vs-Chunks comparisons passed across tiny/huge windows, serial/parallel reads, quoted newlines, comments, empty rows, limits, footers, and typed parse errors.

The bounded-memory claim applies to the structural index. Source bytes and retained output batches still contribute to total memory; this review does not establish an RSS bound or independently verify the throughput figures.


Additional validation: the 100 string-conversion task-budget checks also pass with Julia 1.12.6 and `--threads=1`. Unchanged main `ff0c032` fails three of these checks because chunk conversion uses `ntasks=2` without clamping to the single worker. This PR's `_chunktasks` path fixes that downstream failure too.

